Issue #4, opened 2017-03-10
Installation in a subdirectory?


samwilson:
I'm installing this in a subdirectory for local development, and there are lots of places that are hard-coding absolute URLs. Makes it hard to use `app_dev.php`.

Is it a good idea to change these to use `path()` etc.?

I can send a PR.

samwilson:
See #5

PR #5, opened 2017-03-10
Various chnages to make dev installation easier

samwilson:
* Switch to Symfony 3.2
  (it's backwards compatible to 3.0 which is now not supported)
  to avoid "Twig_Error_Runtime: The "core" extension is not enabled."
* Add `composer validate` to the Travis checks.
* Use `path()` instead of `link()` to construct URLs,
  to account for when the application is loaded via app_dev.php
  or in a subdirectory.
* Give each tool's index route a name that matches that of the tool.
  This helps with constructing these URLs.
* Disable tools (by default) that have not yet been created in this repo.
* Fix license in composer.json and add a description (so the file validates).
* Add phpunit as a dev requirement, for developers who don't have it installed globally.
* Update some documentation.

Bug: GitHub #4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,9 @@ before_install:
 #  - codecov
 
 script:
+    - composer validate
     - composer install
-    - phpunit --coverage-clover=coverage.xml
+    - ./vendor/bin/phpunit --coverage-clover=coverage.xml
     - codecov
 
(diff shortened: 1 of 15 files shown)


MusikAnimal:
Thanks! Tested this out, everything seems to work fine.

PR #6, opened 2017-03-10
Conform to PSR2 coding standards

samwilson:
Add phpcs to Travis to enforce coding standards.
Fix all outstanding deviations from the MediaWiki standards.

Lots of changes here, but nothing substantive. :)

[Work in progress]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
     - composer install
     - ./vendor/bin/phpunit --coverage-clover=coverage.xml
     - codecov
+    - ./vendor/bin/phpcs -s .
 
 sudo: false
 
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -10,6 +10,7 @@ class AppKernel extends Kernel
 {
     public function registerBundles()
     {
+
         $bundles = [
             new Symfony\Bundle\FrameworkBundle\FrameworkBundle(),
             new Symfony\Bundle\SecurityBundle\SecurityBundle(),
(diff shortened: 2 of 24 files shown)


Matthewrbowker:
Pardon my confusion, but we have our own coding standards.  You can read them at  https://xtools.readthedocs.io/en/latest/development.html We do not conform to MediaWiki coding standards in this project as far as I know... @MusikAnimal has that changed?

samwilson:
Sorry! Yes, I should've asked.

So it looks like it's [PSR2](http://www.php-fig.org/psr/psr-2/), is that right? Personally, I think it's slightly better than the MediaWiki style, if only because there's no need to use `mediawiki/mediawiki-codesniffer` (which has some issues).

Do you think the general idea of using phpcs is a good one? I just figured it was worth adding something before too much more code was written. :-)

Matthewrbowker:
@samwilson no worries at all!  

Yes, it does appear to be PSR2, though I've never actually encountered those coding guidelines as written before.  The guidelines i follow were based on coding guidelines I was given in university freshman year.  Thank you for the pointer. 😄 

Yes, I agree that phpcs is a good idea.

samwilson:
@Matthewrbowker cool :-) yeah, I think PSR2 is pretty sensible. Although the MediaWiki sniffs contain some clever stuff, most of it's related to MW core and extension code.

@MusikAnimal do you have an opinion on which CS to use?

MusikAnimal:
No preference, as long as it's consistent. I saw a bunch of different styles in the codebase, some of which looked like mediawiki, so I just went with that. I guess "Read The Docs" is called that for a reason =P

samwilson:
Okay, I've updated this to conform to PSR2. Rather a large PR, I know, but I figure the earlier we put a automatic CS check in place, the less work there is later in fixing code up.

MusikAnimal:
Sorry about the conflicts! If you want you can pull and resolve them, or even reset those two files and I'll update them to PSR2. I'm done for the night so no worries about more conflicts :) Many thanks for setting this up

samwilson:
@MusikAnimal no worries, updated.

PR #13, opened 2017-03-24
Completed Adminstats

Matthewrbowker:
Note: this still has one todo.  I'm not nearly good enough with SQL to debug it though, I've tried my best.  I'll look at it again in a bit.

### DIFF
--- a/src/AppBundle/Controller/AdminStatsController.php
+++ b/src/AppBundle/Controller/AdminStatsController.php
@@ -192,16 +192,16 @@ class AdminStatsController extends Controller
 
         // TODO: Fix this - inactive admins aren't getting shown
         $query = "
-            SELECT user_name, user_id
-            ,SUM(IF( (log_type='delete'  AND log_action != 'restore'),1,0)) AS mdelete
-            ,SUM(IF( (log_type='delete'  AND log_action  = 'restore'),1,0)) AS mrestore
-            ,SUM(IF( (log_type='block'   AND log_action != 'unblock'),1,0)) AS mblock
-            ,SUM(IF( (log_type='block'   AND log_action  = 'unblock'),1,0)) AS munblock
-            ,SUM(IF( (log_type='protect' AND log_action !='unprotect'),1,0)) AS mprotect
-            ,SUM(IF( (log_type='protect' AND log_action  ='unprotect'),1,0)) AS munprotect
-            ,SUM(IF( log_type='rights',1,0)) AS mrights
-            ,SUM(IF( log_type='import',1,0)) AS mimport
-            ,SUM(IF(log_type !='',1,0)) AS mtotal
+            SELECT user_name, user_id,
+                SUM(IF( (log_type='delete'  AND log_action != 'restore'),1,0)) AS mdelete,
+                SUM(IF( (log_type='delete'  AND log_action  = 'restore'),1,0)) AS mrestore,
+                SUM(IF( (log_type='block'   AND log_action != 'unblock'),1,0)) AS mblock,
+                SUM(IF( (log_type='block'   AND log_action  = 'unblock'),1,0)) AS munblock,
+                SUM(IF( (log_type='protect' AND log_action !='unprotect'),1,0)) AS mprotect,
+                SUM(IF( (log_type='protect' AND log_action  ='unprotect'),1,0)) AS munprotect,
+                SUM(IF( log_type='rights',1,0)) AS mrights,
+                SUM(IF( log_type='import',1,0)) AS mimport,
+                SUM(IF(log_type !='',1,0)) AS mtotal
             FROM $loggingTable
             JOIN $userTable ON user_id = log_user
             WHERE  log_timestamp > '$startDb' AND log_timestamp <= '$endDb'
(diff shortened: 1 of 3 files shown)


src/AppBundle/Controller/AdminStatsController.php
```diff
@@ -1,103 +1,264 @@
...
+     * @param string $project Project to run the results against
+     * @param string $start   Date to start on.  Must parse by strtotime.
+     * @param string $end     Date to end on.  Must parse by strtotime.
+     *
      * @Route("/adminstats/{project}/{start}/{end}", name="AdminStatsResult")
+     *
+     * @return Route|\Symfony\Component\HttpFoundation\Response
      */
-    public function resultAction($project, $start = null, $end = null)
+    public function resultAction($project, $start = "1970-1-1", $end = "2038-1-18")
```
samwilson: Is this making the default query be "for all time"? Because it seems to be that on my local install, which is making it go dreadfully slow.

src/AppBundle/Controller/AdminStatsController.php
```diff
@@ -1,103 +1,264 @@
...
+        $userTable = $lh->getTable("user", $dbName);
+        $loggingTable = $lh->getTable("logging", $dbName);
 
-        return $this->render("adminStats/result.html.twig", [
-            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..'),
-            "xtPage" => "adminstats",
-            "xtPageTitle" => "tool_adminstats",
-            "xtSubtitle" => "tool_adminstats_desc",
+        // TODO: Fix this - inactive admins aren't getting shown
+        $query = "
```
samwilson: This query uses a temporary table and filesort, which make it slow. Does it need to join with user? If that's just for getting the username, maybe that can happen after? Although, come to think of it perhaps it needs to group on all fields of an index?

src/AppBundle/Helper/ApiHelper.php
```diff
@@ -27,7 +31,18 @@ private function setUp($project)
     {
         if (!isset($this->api)) {
             $projectInfo = $this->labsHelper->databasePrepare($project);
-            $this->api = MediawikiApi::newFromPage($projectInfo['url']);
+
+            try {
+                if (isset($projectInfo["url"]) && strpos($projectInfo["url"], "api.php") !== false) {
```
samwilson: When is $projectInfo["url"] not set? Either this `isset` can go, or it should also be added to the 'else' clause below (to guard against trying to connect to the API on an unknown URL).

src/AppBundle/Twig/WikiExtension.php
```diff
@@ -69,6 +70,24 @@ public function userLink($username, $projectUrl, $label = null)
...
+     * @return string Markup
+     */
+    public function groupLink($group, $projectUrl, $label = null)
+    {
+        if (!$label) {
+            $label = $group;
+        }
+        // Ignoring this inspection, as we want all of the output on one line.
+        // @codingStandardsIgnoreStart
+        return "<a href='$projectUrl/wiki/index.php?title=Special:ListUsers&group=$group&creationSort=1&limit=50' target='_blank'>$label</a>";
```
samwilson: We shouldn't assume the path to be `/wiki/` (although, I'm not sure how we're handling this; it's a common thing).

Matthewrbowker: We should probably do a pass at all of these API endpoints, following the same logic as the linkWiki() method within Twig... namely check if we're in labs.  I'm open to suggestions though.

MusikAnimal: I may have gotten rid of `linkWiki`, or at least I can't find it now. This was in favour of WikiExtension.php which has helpers for all sorts of links. In there we could add a path helper that fetches and caches the `articlepath` and `scriptpath` from [https://en.wikipedia.org/w/api.php?action=query&meta=siteinfo](meta=siteinfo). I will volunteer to do this since I apparently broke the old system =P

src/AppBundle/Controller/AdminStatsController.php
```diff
@@ -1,103 +1,279 @@
...
+        foreach ($users as $key => $value) {
+            // FIXME: workaround to get the page just to show some data,
+            //  since the API only returns 500 results and the admin returned
+            //  by the query may not have been returned by the API
+            if (empty($data[$value["user_name"]])) {
+                $data[$value["user_name"]] = [
+                    'editcount' => 0,
+                    'groups' => '',
+                ];
+            }
```
MusikAnimal: We need to either use a bot account or use this `massApi` function https://github.com/x-tools/xtools-rebirth/blob/master/src/AppBundle/Helper/ApiHelper.php#L341-L357

For us I think a bot account is preferred, but not sure how we could set that up in such a way to not conflict with the way 3rd wikis are set up. Maybe yet another set of config values in parameters.yml? :)

src/AppBundle/Controller/AdminStatsController.php
MusikAnimal: We do need the usernames since some users are not admins but have made admin actions (we get the current admins with `$api->getAdmins($project);`). Would removing the `ORDER BY` help? My SQL inexperience suggests that might involve a full table scan, when we could do the sorting with PHP.

PR #14, opened 2017-03-31
Add caching helper methods, and caching to API calls

samwilson:
This creates a few methods in a new Helper parent class, to make
it slightly easier to add caching to any helper.
Also some documentation.

Bug: https://phabricator.wikimedia.org/T161057

The specific TTLs here can be changed of course; I'm not sure what they should be. It depends on how people use the tools, but at least some things can be cached for quite a long while (to speed things up).

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -30,13 +30,11 @@ The development server does not cache data.  Any changes you make are visible af
 
 Developing against WMF databases
 --------------------------------
-If you want to use the WMF database replicas, open a tunnel with:
-::
+If you want to use the WMF database replicas, open a tunnel with::
 
     ssh -L 4711:enwiki.labsdb:3306 tools-login.wmflabs.org -N -l your-username-here
 
-And set the following in ``app/config/parameters.yml``:
-::
+And set the following in ``app/config/parameters.yml``::
 
     app.is_labs: 1
     database_replica_host: 127.0.0.1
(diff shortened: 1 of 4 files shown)


src/AppBundle/Helper/ApiHelper.php
```diff
@@ -50,6 +51,7 @@ public function groups($project, $username)
             if (isset($res["batchcomplete"]) && isset($res["query"]["users"][0]["groups"])) {
                 $result = $res["query"]["users"][0]["groups"];
             }
+            $this->cacheSave($cacheKey, $result, 'P7D');
```
MusikAnimal: This info along with global user groups indeed rarely change, but unlike the namespaces this same info may not be requested over and over. Given how popular the tool is, is it worth caching that info in Redis for that long?

src/AppBundle/Helper/EditCounterHelper.php
```diff
@@ -143,6 +140,7 @@ public function getRevisionCounts($userId)
             3
         );
 
+        $this->cacheSave($cacheKey, $revisionCounts, 'P1D');
```
MusikAnimal: This I don't think should be cached at all. Revision counts change frequently and people will probably complain that the numbers are off

MusikAnimal:
Is it possible to cache some things only in development (since I know that would help us)? People will want the live data in production. Frankly I'd say namespaces are the only thing worth caching. User groups and global groups don't change much, but I'm willing to be most of the time the cache will go unused, and they're not especially slow queries.

src/AppBundle/Helper/ApiHelper.php
samwilson: I figured anything that reduces the number of API calls is good, but I'm happy either way. Will remove; good to keep it simple anyway. :)

src/AppBundle/Helper/EditCounterHelper.php
samwilson: Yes, good point.

samwilson:
>Is it possible to cache some things only in development (since I know that would help us)? 

Not sure, but it's a good idea. I think we'd have to specify a different cache adapter wherever we want to do dev-only caching.

I'll pare it back to the minimum for now.

src/AppBundle/Helper/HelperBase.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    protected function cacheHas($key)
+    {
+        /** @var \Symfony\Component\Cache\Adapter\AdapterInterface $cache */
+        $cache = $this->container->get('cache.app');
+        return $cache->getItem($this->getCacheKey($key))->isHit();
+    }
+
+    /**
+     * Get a value from the cache. This is
```
kaldari: "This is"?

docs/development.rst
```diff
@@ -46,7 +44,37 @@ And set the following in ``app/config/parameters.yml``:
...
+=======
+
+Caching should happen in helpers, with appropriate times-to-live.
+
+Every helper should extend HelperBase, which has ``cacheHas()``, ``cacheGet()``, and ``cacheSave()`` methods.
+These should be used in this pattern::
+
+    public function doSomething($input)
+    {
+        $cacheKey = 'ec.something.'.$input;
```
kaldari: What's "ec"? Just curious :)

PR #15, opened 2017-04-05
Edit Counter

samwilson:
This still has a couple of rough edges, but what do you guys think of the general structure?


MusikAnimal:
I can't get this to work. I tried querying for User:Example on enwiki, who has zero edits, and the page never loads :( I actually have to hard reset the server to get the app to be responsive again. What am I doing wrong?

src/AppBundle/Controller/EditCounterController.php
```diff
@@ -6,14 +6,52 @@
...
+        }
+    }
+
+    /**
+     * Every action in this controller should use this to return its view. It will add general 
+     * required parameters.
+     * @param string $name
+     * @param mixed[] $params
+     */
+    protected function view($name, $params)
```
kaldari: Should be able to ditch this function.

PR #16, opened 2017-04-17
Project Mapping

Matthewrbowker:
This is a simple solution to [T161427](https://phabricator.wikimedia.org/T161427), that is; it allows us to map one project name to another.

### DIFF
--- a/app/config/project_map.yml.dist
+++ b/app/config/project_map.yml.dist
@@ -1,4 +1,5 @@
 parameters:
   # Replace <PROJECT_NAME> and <NEW_PROJECT_NAME> with the appropriate values.
-  # One mapping per project, but if you do not map the tables it uses the given project names
+  # You may have multiple mappings per project.
+  # Note that if you do not map, xtools will use the given name.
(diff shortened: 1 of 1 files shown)


app/config/project_map.yml.dist
```diff
@@ -0,0 +1,4 @@
+parameters:
+  # Replace <PROJECT_NAME> and <NEW_PROJECT_NAME> with the appropriate values.
+  # One mapping per project, but if you do not map the tables it uses the given project names
```
samwilson: I'm confused about what this means (oh, and it should say 'projects' rather than 'tables', I think?).

I thought it'd possible to have multiple project names defined on the left hand side, mapping to a actual project names on the right, e.g.:

    app.project.alias1: 'wiki.example.org'
    app.project.alias2: 'wiki.example.org'
    app.project.alias3: 'wiki2.example.org'

Matthewrbowker: Indeed.  My copy-paste job was bad xD I'll make a fix real quick.

PR #17, opened 2017-04-18
Add global xtBaseUrl JS variable and use for API request

samwilson:
I don't think we're assuming top-level directory installation, so I've added this `xtBaseUrl` JS var that can be used to prefix all API requests etc.

(I was getting an error with the NS lookup. Which is a pretty cool feature by the way!)


Matthewrbowker (approved):
Looks good to me.

PR #19, opened 2017-04-19
Bash Tool

Matthewrbowker:
This is the code from the removed tool "bash," I opted to rewrite it rather than try to figure out what the old xTools was doing.


Matthewrbowker:
Thank you for the comments. I'll look at the ID issues Sam raised, and will look at the ui. But as it stands, this is complete.

PR #21, opened 2017-04-21
Hidden params

Matthewrbowker:
This change hides some parameters that are very unlikely to be used on a non WMF labs install.  We can still override them by manually setting them in parameters.yml, however; we don't need to prompt anyone doing a composer install.


samwilson (approved):
Looks good. Nice to make it easier for non-Labs usage.

PR #22, opened 2017-04-24
Automated edits

Matthewrbowker:
This contains code for the (semi-)automated edits tool.


src/AppBundle/Controller/AutomatedEditsController.php
```diff
@@ -0,0 +1,249 @@
...
+    {
+        // Pull the labs helper and check if enabled
+        $lh = $this->get("app.labs_helper");
+        $lh->checkEnabled("autoedits");
+
+        // Pull the values out of the query string.  These values default to
+        // empty strings.
+        $projectQuery = $request->query->get('project');
+        $username = $request->query->get('username');
+        $startDate = $request->query->get('begin');
```
samwilson: The parameter should be 'start'.

src/AppBundle/Controller/AutomatedEditsController.php
```diff
@@ -0,0 +1,249 @@
...
+        $AEBTypes = $this->getParameter("automated_tools");
+
+        // Create a collection of queries that we're going to run.
+        $queries = [];
+
+        $rev = $lh->getTable("revision", $dbName);
+        $arc = $lh->getTable("archive", $dbName);
+  
+        $cond_begin = ( $begin ) ? " AND rev_timestamp > '$begin' " : null;
+        $cond_end = ( $end ) ? " AND rev_timestamp < '$end' ": null;
```
samwilson: These should use named parameters, rather than string construction. Especially as this is end-user input. (I know they're validated to satisfy strtotime above, but I think they should still be parameters.)

src/AppBundle/Controller/AutomatedEditsController.php
```diff
@@ -0,0 +1,249 @@
...
+            'xtPageTitle' => 'tool_autoedits',
+            'xtSubtitle' => 'tool_autoedits_desc',
+            'xtPage' => 'autoedits',
+            'project' => $projectQuery,
+            'namespaces' => $api->namespaces($projectQuery),
+        ]);
+    }
+
+    /**
+     * @Route("/autoedits/{project}/{username}/{begin}/{end}", name="autoeditsResult")
```
samwilson: The 'begin' parameter here is named 'start' in the redirects above.

src/AppBundle/Controller/AutomatedEditsController.php
```diff
@@ -0,0 +1,249 @@
...
+            $cond_end
+        ";
+
+        $cond_begin = str_replace("rev_timestamp", "ar_timestamp", $cond_begin);
+        $cond_end = str_replace("rev_timestamp", "ar_timestamp", $cond_end);
+
+        $queries[] = "
+            SELECT 'deleted' as toolname, count(*) as count
+            from $arc
+            WHERE ar_user_text = '$username'
```
samwilson: Another named parameter.

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
+{% extends 'base.html.twig' %}
+
+{% block body %}
+<div class="col-lg-12">
+<div id="wrapper">
+<main id="content" class="clearfix">
+        <div class="col-lg-8 col-lg-offset-2">
+            </div>
```
MusikAnimal: I don't think you need any of the above elements, since they are in base.html.twig

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+    
+    <div class="panel panel-primary">
+    <header class="panel-heading">
+        <div class="text-center xt-heading-top">
+            <a class="back-to-search" href="{{ path("autoedits") }}"">
+                <span class="glyphicon glyphicon-chevron-left"></span>
+                Back
```
MusikAnimal: `{{ msg('back') }}`

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+            </a>
+            {{ wiki_link("User:" ~ username, url, username) }} &middot; {{ wikiName }}
+        </div>
+    </header>
+    <div class="panel-body xt-panel-body">
+    		<p>
+				<a href="{{ linkWikiScript(url, "title=Special:Log&type=block&user=&year=&month=-1&tagfilter=&user=User:"~username) }} ">{{ msg("block_log") }}</a>
+				{% if isWMFLabs() %}
+				&middot;
+				<a href="//tools.wmflabs.org/guc/?user={{username}}" >Global user contributions</a>
```
MusikAnimal: `{{ msg('global_contribs') }}`, and I personally think it's best if we add `target='_blank'` for external tools, but that's debatable

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+				&middot;
+				<a href="//meta.wikimedia.org/w/index.php?title=Special%3ACentralAuth&target={{username}}" >Global Account Manager</a>
```
MusikAnimal: I don't think we have an i18n message for this, but I would just call it "SUL" or "CentralAuth". We can worry about that later.

Also, you might consider testing for `app.single_wiki` and not `isWMFLabs` here. You should be able to do:
`{{ wiki_link('Special:CentralAuth/'~username, url) }}` since that Special page will work on any wiki

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+        <div class="text-center xt-heading-top">
+            <a class="back-to-search" href="{{ path("autoedits") }}"">
+                <span class="glyphicon glyphicon-chevron-left"></span>
+                Back
+            </a>
+            {{ wiki_link("User:" ~ username, url, username) }} &middot; {{ wikiName }}
+        </div>
+    </header>
+    <div class="panel-body xt-panel-body">
+    		<p>
```
MusikAnimal: `<p class="text-center xt-heading-subtitle">` to match the UI of the other tools, and also the indentation is off throughout the rest of this file

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+				&middot;
+				<a href="//tools.wmflabs.org/guc/?user={{username}}" >Global user contributions</a>
+				&middot;
+				<a href="//meta.wikimedia.org/w/index.php?title=Special%3ACentralAuth&target={{username}}" >Global Account Manager</a>
+				{% endif %}
+			</p>
+
+			<div class="panel panel-default">
+				<div class="panel-heading">
+					<h4  class="topcaption" >Summary <span class="showhide" onclick="javascript:switchShow( 'summary', this )">[hide]</span></h4>
```
MusikAnimal: No inline JavaScript, ever! :) The new system to do this is to use the `macros/layout.html.twig` helpers and let `application.js` do it's thing.

Check out https://github.com/x-tools/xtools-rebirth/blob/master/app/Resources/views/articleInfo/result.html.twig#L594-L628 on how to make a block using the layout helpers. The general format is:
```
{% set content %}
  Your inner content, no .panel-body wrappers or anything
{% endset %}
{{ layout.content_block('i18n-message', content) }}
```

So here `i18n-message` would be `summary`

app/Resources/views/macros/forms.html.twig
```diff
@@ -81,7 +81,7 @@
             <span class="text-muted">({{ msg("optional") }})</span>
         {% endif %}
     </label>
-    <input type="date" class="form-control" name="{{ type }}" placeholder="2016-11-01"{% if required %} required{% endif %} />
+    <input type="date" class="form-control" name="{{ type }}" placeholder="YYYY-MM-DD"{% if required %} required{% endif %} />
```
MusikAnimal: Will `YYYY-MM-DD` make sense in all languages?

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+				</div>
+				<div class="panel-body" id="summary">
+					<table class="leantable table-condensed  xt-table">
+						<tr><td>Starting date</td>		<td>{{ start }}</td></tr>
+						<tr><td>Ending date</td>		<td>{{ end }}</td></tr>	
+						<tr><td>(Semi-)automated edits</td>	<td class="tdnum" >{{ total_semi }}</td></tr>
+						<tr><td>Total edits</td>		<td class="tdnum" >{{ total }}</td></tr>
+						<tr><td>Percentage</td>	<td class="tdnum" >{{ total_pct|number_format(2) }}%</td></tr>
```
MusikAnimal: i18n these if you can :) All of these messages should be there.

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
...
+						<tr><th>{{ msg("tool") }}</th><th>{{ msg("count") }}</th></tr>
+
+			    {% for tool, count in semi_automated %}
+			    	<tr>
+					<td>{{ wiki_link("project:" ~ tool, url, tool) }}</td>
+					<td class="tdnum" >{{ count|number_format(0, ".", ",") }}</td>
+					</tr>
+				{% endfor %}
+			
+					</table>
```
MusikAnimal: See https://github.com/x-tools/xtools-rebirth/blob/master/web/static/js/application.js#L16-L33 on how to build a sortable table. Here it'd be something like (also minding the new class names):
```
<table class="table table-bordered table-hover table-striped">
    {% for key in ['tool', 'count'] %}
        <th class="sort-link sort-link--{{ key }}" data-column="{{ key }}">
            {{ msg(key) | captialize_first }}
        </th>
    {% endfor %}
    {% for tool, count in semi_automated %}
        <tr>
            <td class="sort-entry--tool" data-value="{{ tool }}">
                {{ wiki_link("Project:" ~ tool, url, tool) }}
            </td>
            <td class="sort-entry--count" data-value="{{ count }}">
                {{ count|number_format(0, ".", ",") }}
            </td>
        </tr>
    {% endfor %}
</table>
```

app/Resources/views/macros/forms.html.twig
MusikAnimal: If not we might use `2016-12-31` since with `2016-11-01`, 11 could be interpreted as the day and 01 as the month

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,68 @@
+{% extends 'base.html.twig' %}
```
MusikAnimal: You'll need `{% import 'macros/layout.html.twig' as layout %}` to make the layout helpers work

app/Resources/views/autoEdits/result.html.twig
MusikAnimal: For these unsortable tables (in the summary), we've been doing something like:
```
<div class="col-lg-6 stat-list clearfix">
    <table class="table"><tbody>
        <tr>
            <td>{{ msg('start') }}</td>
            <td>{{ start }}</td>
        </tr>
        ...
    </tbody></table>
</div>
```
`.stat-list` will make it mobile-friendly

app/Resources/views/macros/forms.html.twig
MusikAnimal: Also mind you this generally isn't a problem because modern browsers will show a datepicker for `<input type='date' />`

Matthewrbowker: What I'd really like to do is have it display the actual date that will be submitted... however; I'm not sure that's possible.

MusikAnimal: Actual date that will be submitted? You mean if I pass in the start/end date via URL, it should populate it in the form? We can make that happen

Matthewrbowker: No, I mean the dates the controller uses if no date is given.  For example: https://github.com/x-tools/xtools-rebirth/blob/autoedits/src/AppBundle/Controller/AutomatedEditsController.php#L125

app/Resources/views/autoEdits/result.html.twig
```diff
@@ -0,0 +1,84 @@
...
+{% import 'macros/layout.html.twig' as layout %}
+
+{% block body %}
+    <div class="col-lg-12">
```
MusikAnimal: this shouldn't be needed either. Just `<div class="panel panel-primary">` which frankly might could be moved to base.html.twig as well, but save that for another day

MusikAnimal:
Also looks like you need to pull in master before this can be merged. I assume you'll want the HEAD version of composer.lock, after we upgraded addwiki

MusikAnimal:
Weird, it's actually just the timestamps that are different in composer.lock. Resolve them however you want! I think it was my commit that changed the date format, maybe some weird configuration on my end. Apologies

Matthewrbowker:
Ok I hope I didn't brick that with my latest commit....

PR #24, opened 2017-04-28
Require config option for API path, and use meta=siteinfo

MusikAnimal:
to get metadata about a project, and cache it

Major refactoring of fetching metadata about a project, including namespaces

New 'normalizeProject' to get lang.project.org from various formats

Some i18n fixes

Fun addition to bash quotes

Bug: T163634

### DIFF
--- a/src/AppBundle/Helper/LabsHelper.php
+++ b/src/AppBundle/Helper/LabsHelper.php
@@ -43,8 +43,8 @@ class LabsHelper
     }
 
     /**
-     * Set up LabsHelper::$client and return the database name, wiki name, and URL of a given
-     * project.
+     * Set up LabsHelper::$client and return the database name, wiki name,
+     * and URL of a given project.
      * @todo: Handle failure better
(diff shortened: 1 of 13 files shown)


src/AppBundle/Helper/LabsHelper.php
```diff
@@ -111,6 +75,73 @@ public function databasePrepare($project = 'wiki')
     }
 
     /**
+     * Get the record for the given project in the meta.wiki table
+     * @param  string $project Valid project in the formats:
+     *                         https://en.wikipedia.org, en.wikipedia, enwiki
+     * @return string|false    database record or false if no record was found
```
kaldari: Does this really return a string? Regardless, either the function description or the param description should give a better description of what is actually returned, i.e. what is in this "database record"?

MusikAnimal:
@kaldari I added a little clarification. There are really only 4 values we care about so I listed them.

@samwilson Project class or something similar sounds good. I went in circles trying to keep all the API stuff in the ApiHelper and all the database stuff in LabsHelper.

PR #25, opened 2017-04-29
Update AdminScoreController.php

superyetkin:
Optimized the SQL query to fetch user information

### DIFF
--- a/src/AppBundle/Controller/AdminScoreController.php
+++ b/src/AppBundle/Controller/AdminScoreController.php
@@ -62,9 +62,9 @@ class AdminScoreController extends Controller
 
         $userTable = $lh->getTable("user", $dbName);
         $pageTable = $lh->getTable("page", $dbName);
-        $loggingTable = $lh->getTable("logging", $dbName);
-        $revisionTable = $lh->getTable("revision", $dbName);
-        $archiveTable = $lh->getTable("archive", $dbName);
+        $loggingTable = $lh->getTable("logging_userindex", $dbName);
+        $revisionTable = $lh->getTable("revision_userindex", $dbName);
+        $archiveTable = $lh->getTable("archive_userindex", $dbName);
 
         // MULTIPLIERS (to review)
         $ACCT_AGE_MULT = 1.25;   # 0 if = 365 jours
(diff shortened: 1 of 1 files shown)


src/AppBundle/Controller/AdminScoreController.php
```diff
@@ -62,9 +62,9 @@ public function resultAction($project, $username)
 
         $userTable = $lh->getTable("user", $dbName);
         $pageTable = $lh->getTable("page", $dbName);
-        $loggingTable = $lh->getTable("logging", $dbName);
-        $revisionTable = $lh->getTable("revision", $dbName);
-        $archiveTable = $lh->getTable("archive", $dbName);
+        $loggingTable = $lh->getTable("logging_userindex", $dbName);
+        $revisionTable = $lh->getTable("revision_userindex", $dbName);
+        $archiveTable = $lh->getTable("archive_userindex", $dbName);
```
Matthewrbowker: You just broke backwards compatibility with all versions of MediaWiki.  There's a reason we go through the labs helper - specifically to allow for a table mapping.  

That should be documented better.

src/AppBundle/Controller/AdminScoreController.php
```diff
@@ -105,8 +105,10 @@ public function resultAction($project, $username)
                 AND log_namespace=0
                 AND log_deleted=0 and log_user_text=:username
         UNION
-        SELECT 'blocks' as source, COUNT(*) as value FROM $loggingTable
-            WHERE log_type='block' AND log_action='block' AND log_namespace=2 AND log_deleted=0 AND log_title=:username
+        SELECT 'blocks' as source, COUNT(*) as value FROM $loggingTable l
+            INNER JOIN $userTable u ON l.log_user = u.user_id
+            WHERE l.log_type='block' AND l.log_action='block'
+            AND l.log_namespace=2 AND l.log_deleted=0 AND u.user_name=:username
```
Matthewrbowker: So... you attempt to make this more efficient by adding another Join - which is a slow and heavy process.  That makes absolutely no sense.

superyetkin: Using the "log_user" column in the query makes it better despite adding another join.

Matthewrbowker: Okay, I'm convinced.  Please fix my other comment and we'll give this the thumbs up.  

To clarify my other comment - in order to access a table you use the [MediaWiki table](https://upload.wikimedia.org/wikipedia/commons/9/94/MediaWiki_1.28.0_database_schema.svg) rather than the labs table.  We then map it using app/config/table_map.yml

MusikAnimal: :+1: from me. The new query definitely runs significantly faster, at around 21.73 sec run time with one test I did. The same test took over 3 minutes with the old query, I just ended up killing it.

Indeed Matthew is right however that we have a table mapping system in place. All you need to know is to pass the normal table name to `$lh->getTable`

superyetkin: I am not sure what you mean. Is there any chance you can make the necessary changes?

Matthewrbowker: Sure.  I'll merge then fix it up.  

Thank you for this pull request 😄

PR #26, opened 2017-05-01
A start on more of an OOP structure for Top Edits

samwilson:
This starts to implement the structure we were talking about earlier
today, but of course is very open for discussion!

### DIFF
--- a/app/Resources/views/topedits/index.html.twig
+++ b/app/Resources/views/topedits/index.html.twig
@@ -5,8 +5,8 @@
     <form class="col-lg-8 col-lg-offset-2" action="{{ path("topedits") }}" method="get" accept-charset="utf-8" >
         <fieldset class="form-fieldset">
             {{ forms.user_input() }}
-            {{ forms.wiki_input(project) }}
-            {{ forms.namespace_select(namespaces, true) }}
+            {{ forms.wiki_input(project.domain) }}
+            {{ forms.namespace_select(project.namespaces, true) }}
             {{ forms.page_input(false) }}
         </fieldset>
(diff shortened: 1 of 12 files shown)


MusikAnimal:
Pretty awesome!!! I think this is fine for a start, but you might consider throwing in the `getSiteInfo` code that's currently in ApiHelper. That's where the real metadata comes from, and it will give you the article path and script path, etc. Everywhere you currently see `databasePrepare` (equivalent to `getOne()`) is doing so just to get the dbname, etc., from a inputted project of any valid format. So I'm thinking we could create `normalizeProject` in ProjectRepository (probably static as we talked about), and it calls `getOne` and returns the normalized project name (lang.project.org). Then in Project.php our `getMetadata()` would first get the normalized project name then do what `getSiteInfo` does.

samwilson:
I've moved (duplicated, actually, in order that the updating can be done gradually) the site metadata fetching into ProjectRepository. This is reasonably near ready now.

I think caching is a responsibility of the respositories; what do you think?

And I'm not yet sure how to handle the table-name mutations (i.e. labs or not-labs). Suggestions?

samwilson:
Happy for me to merge this? If so, I'll merge it and then continue expanding the usage of the Project class etc. after.

MusikAnimal:
> Happy for me to merge this? If so, I'll merge it and then continue expanding the usage of the Project class etc. after.

Do it! Were you going to create the Article class? If not I can do so and update ArticleInfo

samwilson:
> Were you going to create the Article class?

Yep. I'll do that now. :)

PR #27, opened 2017-05-01
Fix failing tests that involve namespaces

MusikAnimal:


MusikAnimal:
I don't see how this couldn't fix at least some of those tests...

samwilson:
It seems to be because it's still trying to get project metadata from the `wiki` table even though it's `single_wiki`.

tests/AppBundle/Controller/PagesControllerTest.php
```diff
@@ -23,7 +23,7 @@ public function testIndex()
         $crawler = $client->request('GET', '/pages');
         $this->assertEquals(200, $client->getResponse()->getStatusCode());
 
-        if ($this->container->getParameter('app.is_labs')) {
+        if ($this->container->getParameter('app.is_labs') && !$this->container->getParameter('app.single_wiki')) {
```
samwilson: We should really encapsulate isLabs somewhere else, because it should never be single wiki if it's labs. But this is good for now.

samwilson:
I'm calling this done. :-) Green ticks are good!

MusikAnimal:
@samwilson You're the man! Thank you :)

PR #32, opened 2017-05-15
Table mapping support for one to many

Matthewrbowker:
This code allows us to map different tables based on the labs-specific index.  

It also removes some unnecessary code from the table mapping - an exception that should be handled in the map itself.

Fixes T164556


MusikAnimal:
Looks good!

PR #33, opened 2017-06-01
Fixes for invalid projects throwing 500 errors, and redirect loops ba…

Matthewrbowker:
…sed on default_project.  Also includes other minor fixes.

T166332
T166438

### DIFF
--- a/src/AppBundle/Controller/ArticleInfoController.php
+++ b/src/AppBundle/Controller/ArticleInfoController.php
@@ -11,6 +11,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use AppBundle\Helper\Apihelper;
 use AppBundle\Helper\PageviewsHelper;
 use AppBundle\Helper\AutomatedEditsHelper;
+use Xtools\ProjectRepository;
 
 class ArticleInfoController extends Controller
 {
@@ -76,9 +77,14 @@ class ArticleInfoController extends Controller
         $project = $request->attributes->get('project');
         $page = $request->attributes->get('article');
 
-        // sets the database name within $this->lh
-        $dbValues = $this->lh->databasePrepare($project, 'ArticleInfo');
-        $projectUrl = $dbValues['url'];
+        $projectData = ProjectRepository::getProject($project, $this->container);
+
+        if (!$projectData->exists()) {
+            $this->addFlash("notice", ["invalid-project", $project]);
+            return $this->redirectToRoute("articleInfo");
+        }
+
+        $projectUrl = $projectData->getUrl();
 
         $this->revisionTable = $this->lh->getTable('revision');
 
--- a/src/AppBundle/Controller/AutomatedEditsController.php
+++ b/src/AppBundle/Controller/AutomatedEditsController.php
@@ -6,6 +6,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Xtools\ProjectRepository;
 
 class AutomatedEditsController extends Controller
 {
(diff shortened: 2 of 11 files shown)


src/AppBundle/Controller/ArticleInfoController.php
```diff
@@ -76,9 +77,14 @@ public function resultAction(Request $request)
...
         $page = $request->attributes->get('article');
 
-        // sets the database name within $this->lh
-        $dbValues = $this->lh->databasePrepare($project, 'ArticleInfo');
-        $projectUrl = $dbValues['url'];
+        $projectData = ProjectRepository::getProject($project, $this->container);
+
+        if (!$projectData->exists()) {
+            $this->addFlash("notice", ["invalid-project", $project]);
+            return $this->redirectToRoute("articleInfo");
```
samwilson: I don't think we should redirect in GET requests like this, as it breaks navigation unnecessarily. Instead, if a project doesn't exist, the user should be shown the same form with all their values, as well as an error message, so they can more easily fix whatever typo they made.

src/AppBundle/Controller/ArticleInfoController.php
```diff
@@ -76,9 +77,14 @@ public function resultAction(Request $request)
         $project = $request->attributes->get('project');
         $page = $request->attributes->get('article');
 
-        // sets the database name within $this->lh
-        $dbValues = $this->lh->databasePrepare($project, 'ArticleInfo');
-        $projectUrl = $dbValues['url'];
+        $projectData = ProjectRepository::getProject($project, $this->container);
```
samwilson: Personally, I'd just call this `$project` as that's what it is.

src/AppBundle/Controller/PagesController.php
```diff
@@ -61,20 +62,15 @@ public function indexAction($project = null)
 
         // set default wiki so we can populate the namespace selector
         if (!$project) {
-            $project = $this->container->getParameter('default_project');
+            $project = $this->getParameter('default_project');
```
samwilson: Could also do `$project = ProjectRepository::getDefaultProject($this->container);`

src/AppBundle/Controller/TopEditsController.php
```diff
@@ -65,11 +65,22 @@ public function indexAction(Request $request)
...
+            $namespaces = $project->getNamespaces();
+        }
+
         return $this->render('topedits/index.html.twig', [
             'xtPageTitle' => 'tool-topedits',
             'xtSubtitle' => 'tool-topedits-desc',
             'xtPage' => 'topedits',
-            'project' => $project,
+            'domain' => $domain,
+            'namespaces' => $namespaces,
```
samwilson: In all of these controllers, I don't think it's necessary to pass `domain` and `namespaces` in to the view when `project` is already there and can provide these. Perhaps the form macros need to be more forgiving in what they receive.

src/AppBundle/Controller/ArticleInfoController.php
Matthewrbowker: We really need to get the variable names uniform - in this case, $project is in use elsewhere.

src/AppBundle/Controller/ArticleInfoController.php
Matthewrbowker: Indeed.  I wonder if we can make our forms smart enough that we can post back and have the form display the values that way.  I'm not 100% sure though...

src/AppBundle/Controller/PagesController.php
Matthewrbowker: Huh, didn't know that.  Thanks.

src/AppBundle/Controller/TopEditsController.php
Matthewrbowker: Indeed.  But in this case, it also allows us to set $domain and $namespaces to default values - twig templates don't have very good error handling from what I can tell.

src/AppBundle/Controller/SimpleEditCounterController.php
```diff
@@ -57,11 +58,15 @@ public function resultAction($project, $username)
...
-        $dbName = $dbValues["dbName"];
-        $wikiName = $dbValues["wikiName"];
-        $url = $dbValues["url"];
+        if (!$projectData->exists()) {
+            $this->addFlash("notice", ["invalid-project", $project]);
+            return $this->redirectToRoute("SimpleEditCounter");
+        }
+
+        $dbName = $projectData->getDatabaseName();
+        $url = $projectData->getUrl();
```
MusikAnimal: Speaking of variable names, I think this one should be `$projectUrl` (like it is in some other places), to differentiate from other URLs.

src/AppBundle/Controller/AutomatedEditsController.php
```diff
@@ -67,21 +68,30 @@ public function indexAction(Request $request)
...
+        }
 
-        // TODO: add namespace support
         return $this->render('autoEdits/index.html.twig', [
             'xtPageTitle' => 'tool-autoedits',
             'xtSubtitle' => 'tool-autoedits-desc',
             'xtPage' => 'autoedits',
             'project' => $project,
-            // 'namespaces' => $api->namespaces($project),
+            'namespaces' => $namespaces,
```
MusikAnimal: I don't believe per-namespace support is there yet but I'm going to add it soon, so no harm in uncommenting

Matthewrbowker: I know, I figured it's not a big deal to fix it before it becomes an issue.

Matthewrbowker:
Excellent.  I will leave on my to-do list to go through and redo all of the variable names, it will be a seprate PR though.  I'm merging this one since it has sign-off.

PR #37, opened 2017-06-30
Fix "license" in composer.json: GPL-3.0+ as per README.md

ricordisamoa:


samwilson:
Does the LICENSE file also need to be updated?

ricordisamoa:
> Does the LICENSE file also need to be updated?

Programs released under "version 3 of the License, or (at your option) any later version" usually have version 3 in LICENSE files, just as it is now. However, I actually found a minor issue you may want to fix :sweat_smile: #38

PR #39, opened 2017-07-11
Fix error handling to avoid an XSS error within NoScript

Matthewrbowker:
T170292


samwilson (approved):
Looks good. Needs to be rebased.

MusikAnimal (approved):
I was wondering why the \r\n etc wasn't working! I resolved the conflict using right here on GitHub, hope that's OK. Merging

PR #41, opened 2017-07-12
Add wikitext export to Page History

samwilson:
This adds a new output format to to the top editors table in
the Page History tool.

Bug https://phabricator.wikimedia.org/T170098


MusikAnimal (approved):
:+1: Haven't tested but looks good to me!

PR #42, opened 2017-07-13
Re-enable form when using the browser's back button

MusikAnimal:
Not the prettiest solution but this does the trick. I blame Safari/Firefox and their overzealous caching strategy =P

This PR also cleans up application.scss a bit.


samwilson (approved):
Works in my testing! Looks good.

PR #43, opened 2017-07-13
EditCounter: Rework avg/small/larges and compute only against past 5000 edits

MusikAnimal:
Alternative solution to https://phabricator.wikimedia.org/T170103

I did not write tests because the testable functions do nothing more than reference what the Repo returns.


samwilson (approved):
Looks good.

On the tests: I'd normally say it's worth having tests that even just basically test the names of methods (in effect), because they catch things later when refactoring. But not today, let's get this out! :-)

PR #44, opened 2017-07-13
EditCounter: Remove data on average/smallest/largest edits

MusikAnimal:
So sad... :(

See also the alternative #43


MusikAnimal:
In favour of #43

PR #48, opened 2017-07-17
Treat pages with invalid title as nonexistent

MusikAnimal:
E.g. see http://xtools.wmflabs.org/articleinfo/en.wikipedia.org/2017_NCAA_Division_I_Women%252527s_Basketball_Tournament

The MediaWiki API returns https://en.wikipedia.org/w/api.php?action=query&prop=info|pageprops&titles=2017_NCAA_Division_I_Women%2527s_Basketball_Tournament which does include `missing`, so XTools thinks it's valid.


Matthewrbowker (approved):
Looks good.

PR #50, opened 2017-07-17
AutoEdits: pass in namespace into Result route (whoops)

MusikAnimal:
Ensure required username is provided in TopEdits


Matthewrbowker (approved):
Looks good.

PR #51, opened 2017-07-17
Add 'Times are in UTC' message to the bottom of relevant tools

MusikAnimal:
This is duplicated under the EC's Time card because timezones are more confusing there


samwilson (approved):
Looks good. (I'd have done a paragraph tag, but that's neither here nor there.)

MusikAnimal:
> I'd have done a paragraph tag, but that's neither here nor there

I would too, but Bootstrap likes to put a little bottom padding which doesn't please me =P I could have overridden it, though.

PR #54, opened 2017-07-19
Only show percentages when there's a total

samwilson:
This avoids the division-by-zero errors, and hides the pie charts
when there's no total pie.

Bug: https://phabricator.wikimedia.org/T170608


MusikAnimal (approved):
Looks good!

PR #55, opened 2017-07-20
URL-encode page titles in Twig macros, rm some unneeded unencoding

MusikAnimal:
Twig coding standards


MusikAnimal:
This still isn't ideal because you'll end up with URLs like
`/wiki/User%3AMusikAnimal%2FSubpage` instead of `User:MusikAnimal/subpage` but until we build our own encoder, this seems like the best way to do it.

samwilson (approved):
Looks good to me. I don't think the percent-encoded form for wiki URLs matters too much; most browsers will resolve that for tooltips etc. won't they? Anyway, it'll still work. :-)

PR #56, opened 2017-07-20
Only use the AppBundle for assetic in production

samwilson:
See https://phabricator.wikimedia.org/T170233


Matthewrbowker (approved):
Looks good.

MusikAnimal (approved):
Looks good. We should keep a close eye on prod when this is deployed, though!

PR #57, opened 2017-07-21
Adminscore documentation

Matthewrbowker:
T170892


samwilson (approved):
Magic numbers! :-)

MusikAnimal (approved):
This is a start but we should link to this in the AdminScore interface. The other issue I see is that this isn't translatable :/

Matthewrbowker:
@MusikAnimal That's fine.  I'll explore the option of adding it to the interface proper.

PR #58, opened 2017-07-22
Completed Edit Summary tool

Matthewrbowker:


MusikAnimal (approved):
Looks good! I will merge and see if there's anything I can do on the frontend, but looking at the code it seems to already be inline with the other tools :)

PR #60, opened 2017-07-22
EditSummary: cleanup, add caching and namespace selection

MusikAnimal:
i18n fixes

No tests here... the one db-related function is very specific to this controller and since we're using `fetch()` (and not `fetchAll`) this makes it hard to test.


Matthewrbowker (approved):
Looks good.  Thanks for the look.

PR #62, opened 2017-07-23
Add options to impose rate limiting on a per-session basis

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T168896

### DIFF
--- a/src/AppBundle/EventSubscriber/RateLimitSubscriber.php
+++ b/src/AppBundle/EventSubscriber/RateLimitSubscriber.php
@@ -1,0 +1,150 @@
+<?php
+/**
+ * This file contains only the RateLimitSubscriber class.
+ */
+
+namespace AppBundle\EventSubscriber;
+
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Psr\Cache\CacheItemPoolInterface;
+use Psr\Log\LoggerInterface;
+use DateInterval;
+
+/**
+ * A RateLimitSubscriber checks to see if users are exceeding usage limitations.
+ */
+class RateLimitSubscriber implements EventSubscriberInterface
+{
+
+    /** @var Container The DI container. */
+    protected $container;
+
+    /** @var int Number of requests allowed in time period */
+    protected $rateLimit;
+
+    /** @var int Number of minutes during which $rateLimit requests are permitted */
+    protected $rateDuration;
+
+    /**
+     * Save the container for later use.
+     * @param Container $container The DI container.
+     */
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * Register our interest in the kernel.controller event.
+     * @return string[]
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::CONTROLLER => 'onKernelController',
+        ];
+    }
+
+    /**
+     * Check if the current user has exceeded the configured usage limitations.
+     * @param FilterControllerEvent $event The event.
+     * @throws AccessDeniedHttpException If rate limits have been exceeded.
+     */
+    public function onKernelController(FilterControllerEvent $event)
+    {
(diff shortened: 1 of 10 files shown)


src/AppBundle/EventSubscriber/RateLimitSubscriber.php
```diff
@@ -0,0 +1,95 @@
...
+
+        // Remove reserved characters from URI
+        $uri = preg_replace('/[\{\}\(\)\/\\@:]/', '', $request->getUri());
+
+        $cacheKey = 'ratelimit.'.$sessionId.'.'.$uri;
+        $cache = $this->container->get('cache.app');
+        $cacheItem = $cache->getItem($cacheKey);
+
+        // Initial count, assuming it's their first request with this URI
+        $count = $cacheItem ? (int) $cache->getItem($cacheKey)->get() + 1 : 1;
```
samwilson: When will `$cacheItem` be false? Won't it always be a `CacheItem`? I think this should be `$cacheItem->isHit()` in the first part (and the `$cacheItem` doesn't have to be gotten again in the second part).

samwilson (changes requested):
Looks good! I like the way we can add this without doing anything to the broader application.

I just have a couple of comments, but nothing big.

Matthewrbowker (approved):
Looks good, aside from Sam's comments.

src/AppBundle/EventSubscriber/RateLimitSubscriber.php
```diff
@@ -0,0 +1,123 @@
...
+     * @param FilterControllerEvent $event The event.
+     * @throws AccessDeniedHttpException If rate limits have been exceeded.
+     */
+    public function onKernelController(FilterControllerEvent $event)
+    {
+        $rateLimit = $this->container->getParameter('app.rate_limit_count');
+        $rateDuration = $this->container->getParameter('app.rate_limit_time');
+
+        // Zero values indicate the rate limiting feature should be disabled.
+        if ((int) $rateLimit === 0 || (int) $rateDuration === 0) {
```
MusikAnimal: @samwilson I am typecasting into an int rather than checking with `is_int`. Does this suffice, or is it important to validate that it is actually an integer?

MusikAnimal:
Interesting that Codecov reports 18.7% test coverage introduced with this PR...?

MusikAnimal:
Not done yet... need to get this to log requests that were rejected.

src/AppBundle/EventSubscriber/RateLimitSubscriber.php
samwilson: Yeah, good idea. But I'd move the casting to the initial variable declaration above, because otherwise the other usages below could go awry.

src/AppBundle/EventSubscriber/RateLimitSubscriber.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    private function denyAccess(\Symfony\Component\HttpFoundation\Request $request, $logComment = '')
+    {
+        // Log the denied request
+        $logger = $this->container->get('monolog.logger.rate_limit');
+        $logger->info(
+            "<URI>: " . $request->getRequestUri() .
+            ($logComment != '' ? "\t<Reason>: $logComment" : '') .
+            "\t<User agent>: " . $request->headers->get('User-Agent')
+        );
```
MusikAnimal, line 143: This doesn't come out looking so pretty, but apparently it requires a lot of confusing configuration to get Monolog to allow multi-line log comments. It's still legible which is the important part. See https://stackoverflow.com/questions/29312746/symfony2-how-to-log-multiline-entries-with-monolog

We might also want to add log rotation.

src/AppBundle/EventSubscriber/RateLimitSubscriber.php
MusikAnimal: Done! Good call

samwilson (changes requested):
Looks good! Works for me (I think). The only bug I can see is to move the `administration.rst` doc file up a level, because it's not a tool and so shouldn't be in the `tools` directory; it's not currently being loaded in the ToC.

I reckon log rotation should be done external to the request, so don't worry about it for this feature.

MusikAnimal:
@samwilson Do you have Read the Docs installed locally, or what's the secret there? I'm just going off of other files to figure out to format them, since this apparently is yet another unique flavour of markdown =P

Matthewrbowker:
@MusikAnimal ReadTheDocs uses Sphinx to generate documentation: http://www.sphinx-doc.org/en/stable/ Install sphinx and you should be able to just run "make" in the docs directory.

PR #64, opened 2017-07-25
Add note about pages-created count

samwilson:
Also fixes up some other documentation issues.

Bug: https://phabricator.wikimedia.org/T169955

### DIFF
--- a/app/Resources/views/editCounter/general_stats.html.twig
+++ b/app/Resources/views/editCounter/general_stats.html.twig
@@ -126,12 +126,12 @@
                 <td>{{ ec.averageRevisionsPerPage }}</td>
             </tr>
             <tr>
-                <td>
-                    <a class="glyphicon glyphicon-info-sign tooltipcss" href="https://xtools.readthedocs.io/en/latest/tools/editcounter.html#general-statistics">
-                        <span class="tooltip-body">{{ msg('pages-created-tooltip') }}</span>
+                <td>{% spaceless %}
+                    <a class="glyphicon glyphicon-info-sign tooltipcss tooltipcss--pages-created" href="https://xtools.readthedocs.io/en/latest/tools/editcounter.html#general-statistics">
+                        <div class="tooltip-body">{{ msg('pages-created-tooltip') }}</div>
                     </a>
                     {{ msg('pages-created') }}
-                </td>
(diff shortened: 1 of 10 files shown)


i18n/en.json
```diff
@@ -198,6 +198,7 @@
   "pager-older-n": "{{PLURAL:$1|older 1|older $1}}",
   "pages": "Pages",
   "pages-created": "Pages created",
+  "pages-created-tooltip": "Total of all pages created, inlcuding those created as a redirect when moving a page.",
```
ricordisamoa: typo?

MusikAnimal: qqq? :)

app/Resources/views/editCounter/general_stats.html.twig
```diff
@@ -126,7 +126,12 @@
                 <td>{{ ec.averageRevisionsPerPage }}</td>
             </tr>
             <tr>
-                <td>{{ msg('pages-created') }}</td>
+                <td>
+                    <a class="glyphicon glyphicon-info-sign" target="_blank" title="{{ msg('pages-created-tooltip') }}"
+                       href="https://xtools.readthedocs.io/en/latest/tools/editcounter.html#general-statistics">
+                    </a>
+                    {{ msg('pages-created') }}
+                </td>
```
MusikAnimal: This works but the only way to see the translated message is to hover and _wait_ over the info icon. Most will click on the link which is only in English. If you want we could use our own tooltip system which will make the translatable message more prominent https://github.com/x-tools/xtools/blob/master/app/Resources/views/macros/forms.html.twig#L5-L11

MusikAnimal (changes requested):
needs qqq mainly, but also we might look into using a tooltip rather than using the `title=` attribute

i18n/en.json
samwilson: oops! done.

app/Resources/views/editCounter/general_stats.html.twig
samwilson: I've switched to the `tooltipcss` system, and added a couple of bits to that system; see what you think.

samwilson:
Please re-review.

MusikAnimal:
Good stuff! I add some minor CSS changes

PR #67, opened 2017-07-27
RfX Analysis and RfX Vote Calculator, Round 1

Matthewrbowker:
This is the first round of PR for RfX Analysis and RfX Vote Calculator.  

Note: This is not 100% complete, as I have two pie charts still to insert and minor configuration changes to allow for non-enwiki projects.  However, I would really like to get this code up on the beta sooner rather than later.  

T165709
T165710


src/AppBundle/Controller/RfXVoteCalculatorController.php
```diff
@@ -0,0 +1,196 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\CssSelector\Exception\InternalErrorException;
+use Symfony\Component\Debug\Exception\ContextErrorException;
+use Symfony\Component\HttpFoundation\Request;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Xtools\ProjectRepository;
+use Xtools\RFA;
+use Xtools\User;
+
+// Note: In the legacy xTools, this tool was referred to as "rfap."
```
MusikAnimal: `XTools` :)

src/Xtools/RFA.php
```diff
@@ -0,0 +1,123 @@
...
+
+
+class RFA
+{
+    private $sections;
+    private $data;
+    private $duplicates;
+    private $user_looking_for;
+    private $userSectionFound;
+    private $endDate;
```
MusikAnimal: docs?

src/Xtools/RFA.php
```diff
@@ -0,0 +1,123 @@
...
+    protected function findSig( $input, &$matches ) {
+        //Supports User: and User talk: wikilinks, {{fullurl}}, unsubsted {{unsigned}}, unsubsted {{unsigned2}}, anything that looks like a custom sig template
+        // TODO: Cross-wiki this sucker
+        return preg_match_all(
+            "/\[\[[Uu]ser(?:[\s_][Tt]alk)?\:([^\]\|\/]*)(?:\|[^\]]*)?\]\]" //1: Normal [[User:XX]] and [[User talk:XX]]
+            . "|\{\{(?:[Ff]ullurl\:[Uu]ser(?:[\s_][Tt]alk)?\:|[Uu]nsigned\|)([^\}\|]*)(?:|[\|\}]*)?\}\}" //2: {{fullurl}} and {{unsigned}} templates
+            . "|(?:\{\{)[Uu]ser(?:[\s_][Tt]alk)?\:([^\}\/\|]*)" //3: {{User:XX/sig}} templates
+            . "|\{\{[Uu]nsigned2\|[^\|]*\|([^\}]*)\}\}" //4: {{unsigned2|Date|XX}} templates
+            . "|(?:\[\[)[Uu]ser\:([^\]\/\|]*)\/[Ss]ig[\|\]]/" //5: [[User:XX/sig]] links (compromise measure)
+            , $input, $matches, PREG_OFFSET_CAPTURE
```
MusikAnimal: yeah, maybe move this to `rfx_analysis.yml`?

src/Xtools/RFA.php
```diff
@@ -0,0 +1,186 @@
...
+            . "|\{\{[Uu]nsigned2\|[^\|]*\|([^\}]*)\}\}"
+            //5: [[User:XX/sig]] links (compromise measure)
+            . "|(?:\[\[)[Uu]ser\:([^\]\/\|]*)\/[Ss]ig[\|\]]/";
+
+        return preg_match_all(
+            $regexp,
+            $input,
+            $matches,
+            PREG_OFFSET_CAPTURE
+        );
```
MusikAnimal: In my opinion this function seems fragile (with all that regex) and could benefit from some test cases

src/AppBundle/Controller/RfXAnalysisController.php
```diff
@@ -0,0 +1,213 @@
+<?php
+/**
+ * This file contains the code that powers the RfX Analysis page of xTools.
+ *
+ * @category RfXAnalysis
+ * @package  AppBundle\Controller
+ * @author   Xtools Team <xtools@lists.wikimedia.org>
+ * @license  GPL 3.0
+ * @link     http://xtools.wmflabs.org/rfa
```
MusikAnimal: `XTools` :) And not sure about the mailing list... is anyone checking that? In my opinion `@author` is obsolete anyway now that we have git blame/log

MusikAnimal (changes requested):

src/AppBundle/Controller/RfXAnalysisController.php
Matthewrbowker: I have admin access to the mailing list and check semi-regularly.

src/Xtools/RFA.php
Matthewrbowker: Oh it's quite fragile.  Problem is, with custom signatures this is the least invasive way to do it.  I'd like to ditch the regex though, if you have any ideas I'd love to hear them!

src/AppBundle/Controller/RfXAnalysisController.php
```diff
@@ -0,0 +1,214 @@
...
+use Xtools\RFA;
+
+/**
+ * Class RfXAnalysisController
+ *
+ * @category RfXAnalysis
+ * @package  AppBundle\Controller
+ * @author   XTools Team <xtools@lists.wikimedia.org>
+ * @license  GPL 3.0
+ * @link     http://xtools.wmflabs.org/rfa
```
MusikAnimal: Not sure if the last comment went through... how badly do we want `@author` and `@link`? Author seems redundant, and we have git blame/log which works better anyway. I also wouldn't link to the mailing list which I don't think is being maintained (I resigned my admin rights due to excessive spam). The link makes sense but I question if we want to do that given the app is meant to work on any MediaWiki installation.

MusikAnimal:
How do you feel about squashing the commits?

src/AppBundle/Controller/RfXAnalysisController.php
Matthewrbowker: I put it in because it's required in order to pass Travis.  Maybe `@link` should be the docs?  I maintain the mailing list, it's dead without even spam anymore.

Matthewrbowker:
@MusikAnimal No!!!! I must have credit for all of my work!

Just kidding, squashing is fine 😄

src/AppBundle/Controller/RfXAnalysisController.php
MusikAnimal: Is it required by Travis? I don't see it being used in the docs for other files. `@author` to me doesn't make sense, especially since we're only putting in "XTools Team". Contact info is naturally found on the README, just like the license is in LICENSE, etc. My two cents, but if we put this here we should keep it consistent and repeat it across all the files.

Matthewrbowker: We should keep it consistent.  I've been a little lax in managing docblocks and stuff.  @samwilson you set up the docblock inspection correct?  Is there a way we could ignore the `@author` and `@link` parameters?

samwilson: I think we can remove them. There's nothing in the phpdoc checkstyle thing that highlights them as errors is there, if they're not there? Plenty of classes seem to have no phpdoc tags at all.

src/AppBundle/Controller/RfXVoteCalculatorController.php
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+
+        foreach ($pageTypes as $type) {
+            $type = explode(":", $type, 2)[1];
+
+            $type = str_replace(" ", "_", $type);
+
+            $query = "SELECT DISTINCT p.page_namespace, p.page_title
+FROM `page` p
```
samwilson: I'm getting errors with this because the table names aren't fully qualified and the DB connection is defaulting to meta_p. These table names could be got with something like `$pageTable = $projectData->getRepository()->getTableName($projectData->getDatabaseName(), 'page');`.

src/Xtools/RFA.php
```diff
@@ -0,0 +1,203 @@
...
+                $this->endDate = $matches[1];
+            } elseif ($lastSection != ""
+                && preg_match("/^\s*#?:.*/i", $line) === 0
+            ) {
+                $this->findSig($line, $matches);
+                if (!isset($matches[1][0])) {
+                    continue;
+                }
+                $foundUser = trim($matches[1][0][0]);
+                $this->data[$lastSection][] = $foundUser;
```
samwilson: `$this->data` is never initialized, so if there are no results there's a `Invalid argument supplied for foreach()` down on line 142.

And also: all this in the constructor? Personally I'd move it to another method that's called when required. But whichever; it works this way too! :-)

samwilson (changes requested):
Couple of tiny things. Mostly looks good though!

src/Xtools/ProjectRepository.php
```diff
@@ -206,7 +206,11 @@ public function getMetadata($projectUrl)
             return $this->metadata;
         }
 
-        $api = MediawikiApi::newFromPage($projectUrl);
+        try {
+            $api = MediawikiApi::newFromPage($projectUrl);
+        } catch (\Exception $e) {
+            return null;
```
MusikAnimal, line 213: The doc should be updated to indicate `null` means the API call failed. I would actually instead throw a descriptive exception, because if the API failed we have bigger problems :) Also places that call `getMetadata` in Project.php will error out, such as Project::getTitle

src/AppBundle/Controller/RfXVoteCalculatorController.php
```diff
@@ -0,0 +1,228 @@
+<?php
+/**
+ * This file contains the code that powers the RfX Vote Calculator page of xTools.
+ *
+ * @category RfXVoteCalculator
+ * @package  AppBundle\Controller
+ * @author   XTools Team <xtools@lists.wikimedia.org>
+ * @license  GPL 3.0
+ * @link     http://xtools.wmflabs.org/rfap
```
MusikAnimal: Per above I think we should remove author and link here and in the other files. License is OK but I'd argue it is redundant as the license does not vary file to file within our own code

src/AppBundle/Controller/RfXVoteCalculatorController.php
```diff
@@ -0,0 +1,228 @@
...
+        return 'rfap';
+    }
+
+    /**
+     * Renders the index page for RfXVoteCalculator
+     *
+     * @Route("/rfap", name="rfap")
+     * @Route("/rfap", name="RfXVoteCalculator")
+     *
+     * @return \Symfony\Component\HttpFoundation\Response]
```
MusikAnimal: extra `]`? I think we're just using `Response` in our other controllers, not that it matters

src/AppBundle/Helper/ApiHelper.php
```diff
@@ -315,4 +317,79 @@ private function massApiInternal(&$data)
...
+
+    /**
+     * Query the MediaWiki API and return the text of a single page
+     *
+     * @param xtools/Project $project Project that we're working on
+     * @param string         $title   Title of the page we're pulling
+     *
+     * @return string|null
+     */
+    public function getPageText($project, $title)
```
MusikAnimal: Maybe this should live in `Page`?

src/AppBundle/Helper/ApiHelper.php
```diff
@@ -315,4 +317,79 @@ private function massApiInternal(&$data)
             }
         });
     }
+
+    /**
+     * Query the MediaWiki API and return the text of a single page
+     *
+     * @param xtools/Project $project Project that we're working on
```
MusikAnimal: This implies the `Xtools/Project` class, but we pass it into `$this->setUp` which appears to accept a string

src/AppBundle/Helper/ApiHelper.php
```diff
@@ -315,4 +317,79 @@ private function massApiInternal(&$data)
...
+
+    /**
+     * Query the MediaWiki API and return the text of a many pages
+     *
+     * @param xtools/Project $project Project that we're working on
+     * @param array          $titles   Titles of the pages we're pulling
+     *
+     * @return array|null
+     */
+    public function getMassPageText($project, $titles = [])
```
MusikAnimal: Maybe move this to `PagesRepository`?

src/Xtools/RFA.php
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * @var array Duplicate voters
+     */
+    private $duplicates;
+
+    /**
+     * @var null|string Username of the user we're looking for.
+     */
+    private $user_looking_for;
```
MusikAnimal: camelCase like the other variable names?

src/Xtools/RFA.php
MusikAnimal: I concur, but moreso because I bet we could make this logic testable

MusikAnimal (changes requested):

PR #68, opened 2017-07-27
Fix queries for AdminScore, add missing keys

Matthewrbowker:
T171814


src/AppBundle/Controller/AdminScoreController.php
```diff
@@ -138,16 +138,19 @@ public function resultAction($project, $username)
             WHERE l.log_type='block' AND l.log_action='block'
             AND l.log_namespace=2 AND l.log_deleted=0 AND u.user_name=:username
         UNION
-        SELECT 'afd' AS source, COUNT(*) AS value FROM $revisionTable
-            WHERE rev_page LIKE 'Articles for deletion/%'
-                AND rev_page NOT LIKE 'Articles_for_deletion/Log/%'
-                AND rev_user_text=:username
+        SELECT 'afd' AS source, COUNT(*) AS value FROM $revisionTable r
+          RIGHT JOIN $pageTable p on p.page_id=r.rev_page
```
samwilson: Why `right join` these? Does that mean that we count pages that only have deleted revisions as 1 (i.e. the `rev_page` can be null for some rows where `page_id` is not null)? Oh, except that can't be, because we're then doing `rev_user_text=:username`. So I'm not sure right join does anything here. I'm probably missing something though! :-)

samwilson (commented):
I'm sure it's fine, but I just find right joins to be unusual enough to get attention.

src/AppBundle/Controller/AdminScoreController.php
Matthewrbowker: The Right Join is necessary because the revision table only has a page_id, not page name.

samwilson (approved):

MusikAnimal (approved):
A-OK

PR #70, opened 2017-07-28
Codecov: ignore untestable repositories

MusikAnimal:


MusikAnimal:
I've followed the docs and the `.codecov.yml` should work... but it doesn't here. I'm thinking maybe it needs to be merged into master to take effect. Once we ignore the *Repository.php files our coverage should go up quite a bit.

samwilson:
I think ignoring repositories is fine. They're not supposed to contain any logic, just SQL (which isn't really testable in our set up).

PR #75, opened 2017-08-02
Fix articleinfo API and issues with content type

MusikAnimal:
Better naming for API views

AutoEdits: Wait for DOM to be ready to run chart JS

Upgrade composer packages, but downgrade FOSRestBundle due to bug

Page: rework getPageviews to give sum, handle 404s

Fix tests, add more tests for API


MusikAnimal:
Had to downgrade FOSRestBundle due to https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1702. It's since been fixed but a new stable version hasn't been released. I'll make a note to upgrade when this happens.

Matthewrbowker (approved):
I see no problem with this, even the downgrading of the package.

PR #81, opened 2017-08-17
Page: Don't include ns in title when counting links and redirects.

MusikAnimal:
Add helpers for getNamespaceName and getTitleWithoutNamespace

Bug: https://phabricator.wikimedia.org/T173483

### DIFF
--- a/app/Resources/views/articleInfo/result.html.twig
+++ b/app/Resources/views/articleInfo/result.html.twig
@@ -237,11 +237,12 @@
                         <td>{{ msg('links-external') }}</td>
                         <td>{{ links_ext_count|number_format }}</td>
                     </tr>
-                    <tr>
-                        <td>{{ msg('page-watchers') }}</td>
-                        <td>{{ page.watchers is defined ? page.watchers : '< 30' }}</td>
-                    </tr>
-                    </tr>
+                    {% if page.watchers is not empty %}
+                        <tr>
+                            <td>{{ msg('page-watchers') }}</td>
+                            <td>{{ page.watchers is not empty }}</td>
+                        </tr>
+                    {% endif %}
                     {% if isWMFLabs() %}
                     <tr>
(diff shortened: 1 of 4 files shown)


app/Resources/views/articleInfo/result.html.twig
```diff
@@ -237,10 +237,12 @@
...
-                        <td>{{ msg('page-watchers') }}</td>
-                        <td>{{ page.watchers is defined ? page.watchers : '< 30' }}</td>
-                    </tr>
+                    {% if page.watchers is not empty %}
+                        <tr>
+                            <td>{{ msg('page-watchers') }}</td>
+                            <td>{{ page.watchers is not empty }}</td>
+                        </tr>
+                    {% endif %}
                     </tr>
```
samwilson: Lonely closing tag?

src/Xtools/Page.php
```diff
@@ -58,6 +58,18 @@ public function getTitle()
...
     /**
+     * Get the page's title without the namespace.
+     * @return string
+     */
+    public function getTitleWithoutNamespace()
+    {
+        $info = $this->getPageInfo();
+        $title = isset($info['title']) ? $info['title'] : $this->unnormalizedPageName;
+        $nsName = $this->getNamespaceName();
+        return str_replace(empty($nsName) ? '' : $nsName . ':', '', $title);
```
samwilson: I don't mind it, but I know some people are against `empty()`. In this case, it could probably check for null instead as that's what's returned by `getNamespaceName()`. Although, concatenating null with a string is a string, so maybe this can just be `str_replace($nsName . ':', '', $title);`

samwilson (changes requested):
Looks good. Only the closing tag thing is an issue I think, and that's not even this patch. :-)

samwilson (approved):

PR #84, opened 2017-08-17
Minify JS and version all assets in production

MusikAnimal:
This introduces a Node and npm dependency, apparently there is no
other modern option for minifying JS.

Bug: https://phabricator.wikimedia.org/T172907


samwilson (approved):
Works for me. :)

PR #86, opened 2017-08-22
Check for 'www.' prefix on project names

samwilson:
We were already checking for ".org" suffixes, so this just adds
a "www." prefix as well (and checks for them both as well as
singularly).

Also adds the loading spinner to the Project form field label, to
match what's done for the Namespace form field.

Bug: https://phabricator.wikimedia.org/T170763

### DIFF
--- a/web/static/js/application.js
+++ b/web/static/js/application.js
@@ -346,13 +346,20 @@
             $('#project_input').on('change', function () {
                 var newProject = this.value;
 
+                // Show the spinner.
+                $(this).addClass('show-loader');
+
                 /** global: xtBaseUrl */
                 $.get(xtBaseUrl + 'api/normalizeProject/' + newProject).done(function (data) {
                     // Keep track of project API path for use in page title autocompletion
                     apiPath = data.api;
                     lastProject = newProject;
                     setupAutocompletion();
-                }).fail(revertToValidProject.bind(this, newProject));
+                }).fail(
+                    revertToValidProject.bind(this, newProject)
+                ).always(function () {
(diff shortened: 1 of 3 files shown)


web/static/js/application.js
```diff
@@ -346,13 +346,22 @@
             $('#project_input').on('change', function () {
                 var newProject = this.value;
 
+                // Show the spinner.
+                var $loader = $('span.loader');
+                $('label[for="project_input"]').append($loader);
+                $loader.removeClass('hidden');
```
MusikAnimal: On one hand this makes sense because it's the project field that's triggering the AJAX, but to the user no behaviour changes with that field. I might instead show it on the affected fields like `#user_input` and `#article_input`. Maybe that's even weirder, though... especially for TopEdits, which has a username, namespace and project field! Three loading indicators at once?

Or we could _just_ show the loading indicator on the project field, and disable the affected inputs while the project is loaded. Does that make more sense?

samwilson: Hm, good point. The problem with not showing it while the project is being validated is that when the validation fails they get an error, but it's not till a few seconds later sometimes, and after they've moved to the next field (at which point the project field switches back to what was there before, which is weird I think; doesn't let them correct a typo).

Maybe we could have it spin in the project field, and then do a green tick or red cross depending on what the result is? And maybe move the spinner into the text input itself (or out the other side of it)?

I'm not sure what the best approach is.

Matthewrbowker (approved):

MusikAnimal:
Looks good!

PR #90, opened 2017-08-23
ArticleInfo: add gadget to repo, with endpoint to fetch JS

MusikAnimal:
The generated file will automatically point to the API for the
current XTools installation. You can pass ?uglify=1 to uglify.

### DIFF
--- a/src/AppBundle/Controller/ArticleInfoController.php
+++ b/src/AppBundle/Controller/ArticleInfoController.php
@@ -12,6 +12,8 @@ use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Xtools\ProjectRepository;
 use Xtools\Page;
 use Xtools\PagesRepository;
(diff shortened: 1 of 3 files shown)


MusikAnimal:
Doing it this way because ideally we wouldn't make our script to be WMF-specific. For the same reason I don't think we want the uglified version in our repo either. So here there's just an endpoint at `/articleinfo-gadget.js` that will return working JS for the current XTools installation. We want to sync to the wiki, just browse to `/articleinfo-gadget.js?uglify=1` and copy/paste

src/AppBundle/Controller/ArticleInfoController.php
```diff
@@ -98,6 +99,39 @@ public function indexAction(Request $request)
...
+     */
+    public function gadgetAction(Request $request)
+    {
+        $rendered = $this->renderView('articleInfo/articleinfo.js.twig');
+
+        // SUPER hacky, but it works and is safe.
+        if ($request->query->get('uglify') != '') {
+            $rendered = str_replace('$', '\$', trim($rendered));
+            $rendered = str_replace('"', '\"', trim($rendered));
+            $script = "echo \"$rendered\" | tee temp.js >/dev/null && ";
```
samwilson: This should use a temp file in the sys temp dir, or somewhere. e.g. `$tmpFile = sys_get_temp_dir() . '/xtools_articleinfo_gadget.js';`

src/AppBundle/Controller/ArticleInfoController.php
```diff
@@ -98,6 +99,39 @@ public function indexAction(Request $request)
...
+            $script .= $this->get('kernel')->getRootDir() .
+                '/Resources/node_modules/uglify-es/bin/uglifyjs temp.js --mangle';
+            $script .=  '&& rm temp.js >/dev/null';
+            $process = new Process($script);
+            $process->run();
```
samwilson: Should throw an exception if the process failed, based on `$process->getErrorOutput()`

samwilson (changes requested):
Looks good! The only other thing I'd say other than these two comments is that the uglified version should contain a full URL to the place that generated the ugly, with commented instructions about updating or something. e.g. "This code was automatically generated and should not be edited here manually. For updates, please copy and paste from x."

MusikAnimal:
@samwilson ready for more review when you've time! All excellent ideas, by the way. I like saying it was automatically generated and not to modify it, because people will :)

Matthewrbowker (approved):

samwilson (approved):
Looks absolutely wizard.

PR #91, opened 2017-08-30
EditCounter: use project db name in cache keys of year and month counts

MusikAnimal:


Matthewrbowker (approved):
Looks good, haven't tested it though.

PR #93, opened 2017-09-05
ArticleInfo: fixed average time between edits

MusikAnimal:
Truncate bot list table in addition to editor list table

Copy edit: TOP -> Top


samwilson (approved):
Not tested, but it looks good to me.

PR #95, opened 2017-09-14
Fix to use ProjectRepository to get the replag data.

Matthewrbowker:
This fixes T175796 by using the database name in the query.

### DIFF
--- a/src/AppBundle/Twig/AppExtension.php
+++ b/src/AppBundle/Twig/AppExtension.php
@@ -455,18 +455,18 @@ class AppExtension extends Extension
                 $project = 'enwiki';
             }
 
-            $projectName = ProjectRepository::getProject($project, $this->container)
+            $dbName = ProjectRepository::getProject($project, $this->container)
                 ->getDatabaseName();
 
             $stmt = "SELECT lag FROM `heartbeat_p`.`heartbeat` h
             RIGHT JOIN `meta_p`.`wiki` w ON concat(h.shard, \".labsdb\")=w.slice
-            WHERE dbname LIKE :project OR name LIKE :project OR url LIKE :project LIMIT 1";
+            WHERE dbname LIKE :project LIMIT 1";
 
             $conn = $this->container->get('doctrine')->getManager('replicas')->getConnection();
 
             // Prepare the query and execute
             $resultQuery = $conn->prepare($stmt);
-            $resultQuery->bindParam('project', $projectName);
+            $resultQuery->bindParam('project', $dbName);
             $resultQuery->execute();
 
             if ($resultQuery->errorCode() == 0) {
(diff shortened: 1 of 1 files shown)


src/AppBundle/Twig/AppExtension.php
```diff
@@ -454,6 +455,9 @@ public function replag()
                 $project = 'enwiki';
             }
 
+            $projectName = ProjectRepository::getProject($project, $this->container)
+                ->getDatabaseName();
+
             $stmt = "SELECT lag FROM `heartbeat_p`.`heartbeat` h
             RIGHT JOIN `meta_p`.`wiki` w ON concat(h.shard, \".labsdb\")=w.slice
             WHERE dbname LIKE :project OR name LIKE :project OR url LIKE :project LIMIT 1";
```
samwilson: Does this mean we only need to check against `dbname` here?

src/AppBundle/Twig/AppExtension.php
```diff
@@ -454,6 +455,9 @@ public function replag()
                 $project = 'enwiki';
             }
 
+            $projectName = ProjectRepository::getProject($project, $this->container)
```
samwilson: Could be called `$dbName`?

src/AppBundle/Twig/AppExtension.php
MusikAnimal: That's a good point... we could do the same as we do [`ProjectRepository::getOne`](https://github.com/x-tools/xtools/blob/master/src/Xtools/ProjectRepository.php#L163-L178), but then we're duplicating code. Also above we're re-querying for the project db name, when it's probably already there (at least on result pages, where there's always a `project` in Twig).

If we want to show the replag on index pages in addition to result pages, maybe we should do it like [AutoEditsController](https://github.com/x-tools/xtools/blob/master/src/AppBundle/Controller/AutomatedEditsController.php#L97) where we pass in `project` to Twig, so there will always be a project object (except on the Home and About pages). That way in `base.html.twig` we could call `{% set replagValue = project is defined ? replag(project.dbName) : 0 %}`; or even better, make a `Project::getReplag` function so we only have to put:
```
{% if project is defined and project.replag > getReplagThreshold() %}
    <p class="alert alert-warning xt-alert">
    ...
{% endif %}
```
and we don't need the AppExtension helper.

Matthewrbowker: I like @MusikAnimal 's solution as a long-term solution.  For now, this is a reasonable short-term solution.

MusikAnimal (approved):
This seems fine for now but we might at some point try to utilize existing Project instances.

samwilson (approved):

PR #96, opened 2017-09-21
Add option to make async queries for Edit Counter.

MusikAnimal:
Due to the way curl works, we have to introduce a new parameter
that points to the app's base domain. This also allow us to
optionally use a different domain than the main app server, so
that you can offload resources to a dedicated API server.

Note we still need to document how to forward API requests to another server. See https://phabricator.wikimedia.org/T163284#3628215 for more.

I could not figure out a sane way to add tests for the async features :(

Bug: https://phabricator.wikimedia.org/T163284

### DIFF
--- a/src/AppBundle/Controller/EditCounterController.php
+++ b/src/AppBundle/Controller/EditCounterController.php
@@ -7,9 +7,10 @@ namespace AppBundle\Controller;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
-use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Xtools\EditCounter;
 use Xtools\EditCounterRepository;
 use Xtools\Page;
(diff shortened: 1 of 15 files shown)


src/AppBundle/Controller/EditCounterController.php
```diff
@@ -293,4 +279,114 @@ public function latestglobalAction(Request $request, $project, $username)
             'ec' => $this->editCounter,
         ]);
     }
+
+
+    /**
+     * Below are internal API endpoints for the Edit Counter.
+     * All only respond with JSON and only to requests from the localhost
+     * (see access_control in security.yml).
```
MusikAnimal, line 287: `see access_control in security.yml` -- this isn't actually implemented yet. I want to test this feature in a prod environment first, so I'll use our new prod instance once it's set up.

src/Xtools/Repository.php
```diff
@@ -134,6 +138,29 @@ public function isLabs()
     }
 
     /**
+     * Make a request to the XTools API. This requires that app.multithread.api_url be set.
+     * @param string $endpoint Relative path to endpoint with relevant query parameters.
+     * @param bool $async Set to true to asynchronously query and return a promise.
+     * @return mixed|GuzzleHttp\Promise\Promise
+     */
+    public function queryXTools($endpoint, $async = false)
```
samwilson: I got confused here (not hard to do, certainly). Should this be called `queryXToolsApi()`?

src/Xtools/Repository.php
```diff
@@ -134,6 +138,29 @@ public function isLabs()
...
+    {
+        if (!$this->apiConnection) {
+            $this->apiConnection = $this->container->get('guzzle.client.xtools');
+        }
+
+        // Ensure there's a trailing slash.
+        $apiRoot = trim($this->container->getParameter('app.multithread.api_url'), '/');
```
samwilson, line 162: So am I thinking correctly: I think this should default to the XTools base URL, if `app.multithread.api_url` isn't set (or is set to the default), so that it can still hit the API.

samwilson (commented):
Looks good, works for me in testing! I've just got a couple of stylistic comments, and a query about the new API URL parameter.

samwilson (approved):
Looks good! Works for me.

PR #97, opened 2017-09-27
ArticleInfo: Impose hard limit on number of revisions to analyze

MusikAnimal:
This commit also moves most of the logic out of the
ArticleInfoController and into a model class. Tests added where
feasible to do so.

Bug: https://phabricator.wikimedia.org/T170652


src/Xtools/ArticleInfo.php
```diff
@@ -0,0 +1,989 @@
...
+
+    /**
+     * Get the page assessments of the page.
+     * @see https://www.mediawiki.org/wiki/Extension:PageAssessments
+     * @return string[]
+     */
+    public function getAssessments()
+    {
+        if (!is_array($this->assessments)) {
+            $this->assessments = $this->page->getAssessments();
```
samwilson, line 346: I'm getting a Twig error when requesting `/articleinfo/en.wikisource.org/Main_Page`: `Impossible to access an attribute ("assessment") on a boolean variable ("").` because `Page::getAssessments()` can also return false.

samwilson (commented):
This works in my testing of everything apart from this one thing. (Of course, I'm not sure I'm testing comprehensively.)

src/Xtools/ArticleInfo.php
MusikAnimal: Yeah I think this is actually breaking in production too (silently failing). Might as well fix it with this PR, I'm on it!

Matthewrbowker (approved):
Looks good.  I haven't tested it on my system but I trust you did :)

MusikAnimal:
Merging... hope that's OK! I will do a 3rd re-do of ArticleInfo at some point. Believe it or not `parseHistory` is lightyears better than it used to be :)

PR #98, opened 2017-09-27
EditCounter: show longest block as the duration, not expiry

MusikAnimal:
Loop through the block log to get the amount of time the user
was blocked.

Bug: https://phabricator.wikimedia.org/T174012

### DIFF
--- a/src/Xtools/EditCounter.php
+++ b/src/Xtools/EditCounter.php
@@ -64,11 +64,11 @@ class EditCounter extends Model
     protected $editSizeData;
 
     /**
-     * Duration of the longest block in days; -1 if indefinite,
+     * Duration of the longest block in seconds; -1 if indefinite,
      *   or false if could not be parsed from log params
      * @var int|bool
      */
-    protected $longestBlockDays;
+    protected $longestBlockSeconds;
 
     /**
      * EditCounter constructor.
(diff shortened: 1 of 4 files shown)


Matthewrbowker (approved):

src/Xtools/EditCounter.php
```diff
@@ -308,49 +317,117 @@ public function countBlocksReceived()
...
+         */
+        $lastBlock = null;
+
+        foreach ($blocks as $index => $block) {
+            list($timestamp, $duration) = $this->parseBlockLogEntry($block);
+
+            if ($block['log_action'] === 'block') {
+                // This is a new block, so first see if the duration of the last
+                // block exceeded our longest duration. -1 duration means indefinite.
+                if ($lastBlock[1] > $this->longestBlockSeconds || $lastBlock[1] === -1) {
```
kaldari: The first time this loop executes, it's going to throw an undefined variable notice about `$lastBlock[1]`. Shouldn't `$lastBlock = [$timestamp, $duration];` happen earlier?

kaldari (changes requested):

src/Xtools/EditCounter.php
MusikAnimal: We don't want to set the timestamp and duration yet, but you're right the initial declaration should be like `$lastBlock = [null, null]`

MusikAnimal:
Code review concerns addressed; going to go ahead and merge so I can get this and [T177172](https://phabricator.wikimedia.org/T177172) deployed, hope that's OK

PR #100, opened 2017-10-03
Impose hard limit on number of edits to process

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T170652


MusikAnimal:
Bolding going to merge. This isn't the most ideal way to do this (instead we should be doing `EXPLAIN`s to see how many rows are going to be scanned), but this still works as a temporary safeguard. Sorry I couldn't write better tests!

PR #102, opened 2017-10-05
Get the page title without namespace when generating Top Edits path

Matthewrbowker:
Bug: T177300


samwilson (approved):
Makes sense :)

MusikAnimal (approved):
Easy peasy

PR #103, opened 2017-10-05
Specify target platform in Composer 

samwilson:
and remove now-redundant requirement statements

(maybe this will work?)


MusikAnimal:
I got prod02 to pull this in and run composer, and no modifications were made to the symfony requirement files -- so hopefully this fixed it! Going to merge and do more testing just to make sure the new versions of the packages don't break anything.

PR #105, opened 2017-10-10
EditCounter: Make bars of month and year charts consistent

MusikAnimal:
Remove legends from year/month charts, instead using the namespace table
to toggle namespaces. When toggled, update the percentages in the table
and show/hide the dataset from the year and month charts.

Bug: https://phabricator.wikimedia.org/T174581


MusikAnimal:
Same here, just going to merge. I used the PR only to get Scrutinizer results. This fix is somewhat urgent too so people will stop complaining about how "ugly" the new Edit Counter is =P

PR #107, opened 2017-10-11
Deletion Counter

Matthewrbowker:
Add the Deletion Counter tool


Matthewrbowker:
This one is going to require some discussion...

This pull request introduces a new tool, called the Deletion Counter.  Its intended purpose is to break down which process an administrator used for deleting an edit.  It was originally inspired by https://www.mediawiki.org/wiki/Topic:Tvegxdd0obtx028l, which at the time was found too en-wiki-centric.  However, much like or RfX tools I've implemented it to use YAML files so it's flexible enough to support any wiki.  All we would need is wikis to file Phabricator tasks so we know what's what.

MusikAnimal:
I like it! FYI SoWhy did end up creating his own tool :-/ https://tools.wmflabs.org/sowhy/delanalyze.php

I'd hate to try to 1-up, but we've already put the work into it, and the way you've designed it indeed seems XTools-worthy. The YAML approach is definitely appropriate but if there is no configuration, we might consider parsing [[MediaWiki:Deletereason-dropdown]] as it would then _just work_ ™ for all wikis.

The other thing is ideally we could (optionally) give it a second layer of configuration in the YAML file like:
```YAML
parameters:
  deletion_counter:
    enwiki:
      speedy:
        G1: "WP:CSD#G1"
        G2: "WP:CSD#G2"
        ...
    XFD:
      AFD: "Articles for deletion"
      TFD: "Templates for discussion"
      MFD: "Miscellany for deletion"
    PROD:
      PROD: "WP:PROD"
      BLPPROD: "WP:BLPPROD"
```
This way we could get an overall comparison of the speedy vs XfD vs PROD, in addition to a breakdown of the individual types of speedy, XfD and PROD.

Also, if possible maybe we could write a test or two. I think we should aim to have the overall test coverage change by >= 0% with each commit/PR. I hope to eventually be able to put the Codecov badge on the README :) Right now it's at an embarrassing 28%!

src/AppBundle/Controller/DeleteCounterController.php
```diff
@@ -0,0 +1,198 @@
...
+         */
+        $project = ProjectRepository::getProject($project, $this->container);
+        $projectRepo = $project->getRepository();
+
+        $user = UserRepository::getUser($username, $this->container);
+
+        $userID = $user->getId($project);
+
+        if (!$project->exists()) {
+            $this->addFlash('notice', ['invalid-project', $project]);
```
MusikAnimal: This one seems to have merit: `$project` at this point is a `Project` and not a string as `addFlash` expects

src/AppBundle/Controller/DeleteCounterController.php
```diff
@@ -0,0 +1,198 @@
...
+            $this->addFlash('notice', [ 'no-result', $username ]);
+            return $this->redirectToRoute(
+                'DeletionCounter',
+                [
+                    'project' => $project->getDomain()
+                ]
+            );
+        }
+
+        $data = $types[$project->getDatabaseName()];
```
MusikAnimal: You might reuse `$dbName` here instead of calling `$project->getDatabaseName()` again, but personally I think the config should go by domain to be consistent with semi_automated.yml and rfx_analysis.yml.

src/AppBundle/Controller/DeleteCounterController.php
```diff
@@ -0,0 +1,198 @@
...
+                [
+                    'project' => $project->getDomain()
+                ]
+            );
+        }
+
+        $total = 0;
+
+        foreach ($resultQuery->fetchAll() as $row) {
+            $results[$row["type"]] = $row["count"];
```
MusikAnimal: I would agree with this one too (putting `$results = []` somewhere around line 165). Otherwise it would seemingly error out here and on line 174 if `$resultQuery->fetchAll()` doesn't return anything.

app/Resources/views/deletionCounter/result.html.twig
```diff
@@ -0,0 +1,62 @@
...
+                <tr>
+                    <th>{{ msg("deletion-process") }}</th>
+                    <th>{{ msg("count") }}</th>
+                </tr>
+                </thead>
+                <tbody>
+                {% for key, value in results %}
+                    <tr>
+                        <td>{{ key }}</td>
+                        <td>{{ value }}</td>
```
MusikAnimal: `{{ value|format_number }}`

app/Resources/views/deletionCounter/result.html.twig
```diff
@@ -0,0 +1,62 @@
...
+                    </tr>
+                {% endfor %}
+                <tr>
+                    <td>
+                        <b>{{ msg("total") }}</b>
+                    </td>
+                    <td>
+                        <b>{{ total }}</b>
```
MusikAnimal: `{{ total|format_number }}`

app/config/deletion_counter.yml
```diff
@@ -0,0 +1,8 @@
+parameters:
+  deletion_counter:
+    enwiki:
```
MusikAnimal: I think this should go by domain instead of database name, to be consistent with semi_automated.yml and rfx_analysis.yml

MusikAnimal (changes requested):
Looks good! My comments around the two-layer config (so we can compare total speedys vs XfDs vs PRODs) is just a thought and can certainly by a v2 thing :) Same with parsing MediaWiki:Deletereason-dropdown, which might be really difficult.

app/config/deletion_counter.yml
MusikAnimal: I think the config should maybe go by domain to be consistent with semi_automated.yml and rfx_analysis.yml

app/Resources/views/deletionCounter/result.html.twig
MusikAnimal: `{{ value|number_format }}`

app/Resources/views/deletionCounter/result.html.twig
MusikAnimal: `{{ total|number_format }}`

MusikAnimal (changes requested):
Looks good! My comments about the two-layer config (so we can compare overall speedy vs XfD vs PROD) is more of a thought, and certainly can be a v2 thing :) Same with parsing [[MediaWiki:Deletereason-dropdown]], which might actually be pretty challenging.

MusikAnimal (changes requested):
Looks good! My comments about the two-layer config (so we can compare overall speedy vs XfD vs PROD) is more of a thought, and certainly can be a v2 thing :) Same with parsing [[MediaWiki:Deletereason-dropdown]], which might turn out to be fairly challenging.

MusikAnimal (changes requested):
Looks good! My comments about the two-layer config (so we can compare overall speedy vs XfD vs PROD) is more of a thought, and certainly can be a v2 thing :) Same with parsing [[MediaWiki:Deletereason-dropdown]], which might turn out to be fairly challenging.

MusikAnimal (changes requested):
Looks good! My comments about the two-layer config (so we can compare overall speedy vs XfD vs PROD) is more of a thought, and certainly can be a v2 thing :) Same with parsing [[MediaWiki:Deletereason-dropdown]], which might turn out to be fairly challenging.

app/Resources/views/deletionCounter/result.html.twig
```diff
@@ -0,0 +1,62 @@
...
+
+    <div class="panel-body xt-panel-body">
+        {{ wiki.userLinks(user, project, xtPage) }}
+
+        {% set content %}
+            <div class="col-lg-6">
+            <table class="table table-bordered table-hover table-striped xt-show-hide--target">
+                <thead>
+                <tr>
+                    <th>{{ msg("deletion-process") }}</th>
```
MusikAnimal: Before we commit this into version control and the translators get to work, we might consider just using the term "type" for now. I say this because eventually it'd be nice if we had a breakdown of the types of deletions for a given process (A1, A2, etc.). For now "type" is still a fitting word for this column and we already have a message for it.

app/Resources/views/deletionCounter/result.html.twig
```diff
@@ -0,0 +1,62 @@
...
+{% block body %}
+<div class="panel panel-primary">
+    {{ layout.userHeading(user, project, xtPage) }}
+
+    <div class="panel-body xt-panel-body">
+        {{ wiki.userLinks(user, project, xtPage) }}
+
+        {% set content %}
+            <div class="col-lg-6">
+            <table class="table table-bordered table-hover table-striped xt-show-hide--target">
```
MusikAnimal: `xt-show-hide--target` is for adding a [show]/[hide] links for subsections within a `panel`. Here we have just the one so this class can be removed. I still need to document this, sorry about that :)

src/AppBundle/Controller/DeleteCounterController.php
```diff
@@ -0,0 +1,211 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Xtools\Project;
+use Xtools\ProjectRepository;
+use Xtools\User;
+use Xtools\UserRepository;
+
+/**
+ * This controller handles the Simple Edit Counter tool.
```
MusikAnimal: `Delete Counter`

src/AppBundle/Controller/DeleteCounterController.php
```diff
@@ -0,0 +1,211 @@
...
+     *
+     * @return string
+     */
+    public function getToolShortname()
+    {
+        return 'dc';
+    }
+
+    /**
+     * The Simple Edit Counter search form.
```
MusikAnimal: `Delete Counter`

src/AppBundle/Controller/DeleteCounterController.php
MusikAnimal: Or `Deletion Counter`, whatever we're calling it :)

app/Resources/views/deletionCounter/result.html.twig
```diff
@@ -0,0 +1,62 @@
...
+                }} ) %}
+            {% endfor %}
+                <div style="clear:both"></div>
+            <div class="basic-info-charts col-lg-12 clearfix">
+        {{ chart.pie_chart('percent', code) }}
+            </div>
+            {% endif %}
+                {% endset %}
+            {{ layout.content_block('tool-dc-desc', content) }}
+        </div>
```
MusikAnimal: Indentation seems off within the `.xt-panel-body` block (child tags should be indented below the parent, same with [Twig](https://twig.symfony.com/doc/2.x/coding_standards.html)). It also looks like maybe the `{% endset %}` on line 57 is supposed to go here below the closing `</div>` on line 59?

MusikAnimal:
Left a few more comments. After those are addressed, I think it's OK to merge, but it'd probably be nice to wait a while before enabling this tool. First, I think we could offer more features, but moreover I'd feel really bad turning this on when I _just_ told SoWhy we weren't doing it. We should let his tool shine for a while :) I also think in general we should avoid adding anymore tools that have to be configured per-wiki, or at least discuss it prior to development. To me, XTools is about MediaWiki and should work out of the box. That being said this one has potential because again we might be able to parse MediaWiki:Deletereason-dropdown, which would be really cool!

Matthewrbowker:
Agreed.  I have on my todo list to parse the MediaWiki:Deletereason-dropdown, it's perfectly fine to leave it off for now if it is merged.

Matthewrbowker:
Obvious this is not going to get merged at this time.  Closing.

MusikAnimal:
Just don't delete the branch yet! Once the RfX tools are up to scale (which I've actually done a little work on), I think we could add those and the deletion counter under a "More" menu.

Matthewrbowker:
I'm ok with that, we can leave the branch intact just fine.

PR #111, opened 2017-10-12
Abstract XtoolsController for refactoring code in other controllers

MusikAnimal:
This is crazy big... sorry. Mostly refactoring.


MusikAnimal:
Sorry for the yet another giant PR... What's happening here is I'm moving a lot of the repetition across the controllers to an abstract XtoolsController, and adding things like date validations and URL decoding (which might fix https://phabricator.wikimedia.org/T176030#3663391). That's the bulk of it, but other changes have been added like showing "no contributions found" in the view rather than redirecting back to the form -- which is what almost all of the changes are to the Twig templates. It looks crazier than it is, promise!

This PR also includes other various code quality improvements, mostly suggested by Scrutinizer.

src/AppBundle/Controller/XtoolsController.php
```diff
@@ -0,0 +1,323 @@
...
+use Xtools\PagesRepository;
+
+/**
+ * XtoolsController supplies a variety of methods around parsing and validing
+ * parameters, and initializing Project/User instances. These are used in
+ * other controllers in the AppBundle\Controller namespace.
+ */
+abstract class XtoolsController extends Controller
+{
+    const LANGUAGELESS_PROJECTS = [
```
samwilson: Should this list go in a config file?

samwilson (approved):
Looks good, I haven't extensively tested but the general vibe is one of goodness. :-)

I like the canonical list of query string params. Every app should have that!

I don't think you need the `use` statements for the XtoolsController because it's in the same namespace.

MusikAnimal:
CR concerns addressed, so going to merge. Hope this is OK! From my testing this indeed fixes https://phabricator.wikimedia.org/T176030#3663391 which we were aiming to have resolved by dewiki's announcement on Monday.

PR #128, opened 2017-10-30
Tagging Interface

Matthewrbowker:
This generates a dev-only tag interface located at /meta/tag.  

This will work when you run `php bin/console server:run` under your user - there may be some permission issues on Linux if there is a user mismatch.  

Note that I have not tested the Git component of this tool.


MusikAnimal:
Neat! This to me seems more appropriate as a command-line tool, but I suppose that might get tricky when entering multi-line release notes. I know both Sam and I use Apache instead of the built-in Symfony server, so I wonder if that will cause permission errors. I'll test it out soon!

Just need those codesniffer tests to pass, and you can add `@codeCoverageIgnore` to the bottom of the `tagIndexAction` doc.

src/AppBundle/Controller/MetaController.php
```diff
@@ -186,4 +204,104 @@ public function recordUsage(Request $request, $tool, $project, $token)
...
+     * @Route("/meta/tag", name="MetaTag")
+     *
+     * @return Response
+     */
+    public function tagIndexAction()
+    {
+        if ($this->container->getParameter("kernel.environment") !== "dev") {
+            // We don't want any more than this, this must be hidden entirely on
+            // prod environments.
+            throw new FileNotFoundException();
```
MusikAnimal: This might should be `throw $this->createNotFoundException('This action is only available in the dev environment.');` (or something)

Matthewrbowker:
I opted for the web interface for two reasons
1. I found myself loading much of Symfony's environment or attempting to parse YML to get information (That was a bigger deal in an earlier implementation, now I just need the path to the kernel and the current version)
2. Multi-line release notes were impossible on command line without just throwing you into vim.

samwilson:
Yeah, I get `Warning: file_put_contents(.../xtools/RELEASE_NOTES.md): failed to open stream: Permission denied` but that's not unfixable.

Overall though, I'd love this sort of tool to be available for all projects! Do you think it'd be much work to generalize it as a project-agnostic PHP tool, installable globally with Composer? What XTools-specific assumptions are made?

samwilson:
Could we use this sort of thing? https://github.com/liip/RMT

Matthewrbowker:
The only real assumption I make is the locations of files.  And how to load the current version string...

We could, haven't explored it yet.  Looks like it shouldn't be too hard to configure.

Matthewrbowker:
After looking into RMT, this looks like a wonderfully useful tool.  I'm going to close this and attempt to set up RMT and open a new PR.  Apologies about the delay.

PR #134, opened 2017-11-24
Several improvements to adminscore

Matthewrbowker:
T179508
T179764
T179763


app/Resources/views/adminscore/result.html.twig
```diff
@@ -20,14 +20,21 @@
                                 <td>
                                     <span class="text-success">{{ master[key].score|diff_format }}</span>
                                 </td>
+                                <td>
+                                    <span class="text-muted small">
+                                        {{ msg(key~"-desc", [master[key].value, master[key].value|number_format]) }}
+                                    </span>
+                                </td>
```
MusikAnimal, line 27: With the extra column, the text is getting wrapped. Changing `stat-list` div to have `col-lg-12` instead of 6 should do it

This also doesn't show up very well on mobile, but not a big deal. I can't think of a straightforward solution off of the top of my head.

i18n/en.json
```diff
@@ -48,6 +51,7 @@
 	"block-log": "Block log",
 	"block-longest": "Longest block",
 	"blocks": "Blocks",
+	"blocks-desc": "$2 {{PLURAL:$1|block|blocks}} administered",
```
MusikAnimal, line 54: Just curious, why do we include this in the score? I thought Admin Score was more for non-admins. Even for admins, it seems odd to only account for blocks and not other admin actions.

MusikAnimal (changes requested):
Good stuff! Just a few comments, and looks like there are some codesniffer bugs.

i18n/en.json
Matthewrbowker: I'm not quite sure, our AdminScore was copied from the old one entirely.  I would be OK removing it, maybe we should open a Phabricator task to discuss it?

MusikAnimal: Nah I think it's reasonable to just remove it. Doesn't have to be part of this PR, though!

MusikAnimal (approved):

PR #142, opened 2017-12-14
EditCounter: add section listing legible rights changes

MusikAnimal:
This adds a "user rights" section to the Edit Counter, which simply lists rights changes to the account over time, in a legible format (as opposed to the [MediaWiki rights log](https://en.wikipedia.org/w/index.php?title=Special:Log&user=&page=User%3AL235&type=rights])). The section is only shown if they have had user rights changes. The underlying query to the logging table is quite fast (< 40ms), even for accounts with a high edit count.

I have no idea why the build is failing on Scrutinizer. For some reason the git sha there isn't the correct length. Build is passing on Travis, so I'm going to assume everything is OK...


Matthewrbowker (approved):
Looks good to me.

PR #143, opened 2017-12-15
Add options for deleted pages to Pages Created tool

Albert221:
Bug: T172003


src/AppBundle/Controller/PagesController.php
```diff
@@ -64,21 +64,27 @@ public function indexAction(Request $request)
...
-     * @Route("/pages/{project}/{username}/{namespace}/{redirects}/{offset}", name="PagesResult")
+     * @Route("/pages/{project}/{username}/{namespace}/{redirects}/{deletedpages}/{offset}", name="PagesResult")
      * @param Request $request
      * @param string|int $namespace The ID of the namespace, or 'all' for all namespaces.
      * @param string $redirects Whether to follow redirects or not.
+     * @param string $deletedpages Whether to include deleted pages or not.
      * @param int $offset Which page of results to show, when the results are so large they are paginated.
-     * @return RedirectResponse|Response
-     * @codeCoverageIgnore
+     * @return RedirectResponse|Response* @codeCoverageIgnore
```
MusikAnimal: Needs a line break before `* @codeCoverageIgnore`

MusikAnimal (changes requested):
Looking good! :)

I probably should have given a little background, in case you weren't familiar. For revision information, there are two tables. `revision` is for live revisions, and `archive` is for deleted. So here instead of filtering on `type`, we could conditionally use the subqueries. E.g. for "exclude deleted pages", you'd omit this part of the query: https://github.com/x-tools/xtools/blob/1de48dd3377caecd9be918beeb4a62b6b38a819d/src/Xtools/UserRepository.php#L387-L403

Taking out the unneeded subqueries should speed it up quite a bit. You might need to update [getNamespaceAndRedirectConditions](https://github.com/x-tools/xtools/blob/1de48dd3377caecd9be918beeb4a62b6b38a819d/src/Xtools/UserRepository.php#L312-L321) and [getUserConditions](https://github.com/x-tools/xtools/blob/1de48dd3377caecd9be918beeb4a62b6b38a819d/src/Xtools/UserRepository.php#L344-L353) (I'm not certain). I realize this probably makes it a lot more complicated... so let me know if you need help!

To avoid confusion, I think we want to hide the [deleted column](https://github.com/x-tools/xtools/blob/1de48dd3377caecd9be918beeb4a62b6b38a819d/app/Resources/views/pages/result.html.twig#L93-L96) from the summary in the view, when set to "only include deleted pages".

Finally, it looks like the phpDoc blocks [need to be updated](https://travis-ci.org/x-tools/xtools/jobs/317051942#L781-L784). The test failure in Scrutinizer can be ignored. I fixed that and it should pass with the next build.

src/Xtools/Pages.php
```diff
@@ -29,6 +29,9 @@ class Pages extends Model
     /** @var string One of 'noredirects', 'onlyredirects' or blank for both. */
     protected $redirects;
 
+    /** @var string One of 'both', 'live' or 'deleted' */
+    protected $deletedPages;
```
MusikAnimal: I would probably just name this property `deleted`, since it doesn't actually contain the pages.

MusikAnimal (changes requested):

Albert221:
Thank you for very detailed CR! I've fixed everything.

src/Xtools/UserRepository.php
```diff
@@ -304,15 +308,17 @@ public function getPagesCreated(
      * Get SQL fragments for the namespace and redirects,
      * to be used in self::getPagesCreatedInnerSql().
      * @param  string|int $namespace Namespace ID or 'all'.
-     * @param  string $redirects     One of 'noredirects', 'onlyredirects' or blank for both.
+     * @param  string $redirects One of 'noredirects', 'onlyredirects' or blank for both.
+     * @param  string $deleted One of 'live', 'deleted' or 'both'.
      * @return string[] With keys 'namespaceRev', 'namespaceArc' and 'redirects'
```
MusikAnimal: We should add `'deleted'` to this list

MusikAnimal (changes requested):
The new SQL looks good! :)

I'm still trying to figure it out, but it it looks like the option "noredirects" is now returning redirects, when it shouldn't. Then I get an error when I select "only redirects".

Possibly related, I think we are missing some logic in [Pages::getCounts()](https://github.com/x-tools/xtools/pull/143/files#diff-16a388c07d4e5dd367839c9cd32e84c0R237) that controls when `deleted` key/value is added, based what value `$this->deleted` is set to.

Bonus points (might be better for a separate PR): Updating the [countPagesApiAction](https://github.com/x-tools/xtools/pull/143/files#diff-a0aafe0b72a62b5ba221428cd7375fc6R160) and [getPagesApiAction](https://github.com/x-tools/xtools/pull/143/files#diff-a0aafe0b72a62b5ba221428cd7375fc6R212) endpoints to use the deleted parameter. The [documentation](http://xtools.readthedocs.io/en/latest/api/user.html#number-of-pages-created) would also [need to be updated](http://xtoolsproject.org/en/stable/development.html#writing-the-docs).

I'm going to check back in on this early tomorrow, and try to help as much as I can.

Albert221:
Above commit is still WIP, I didn't commit updated documentation yet.

> I'm still trying to figure it out, but it it looks like the option "noredirects" is now returning redirects, when it shouldn't. Then I get an error when I select "only redirects".

It happens on `master` too, I double checked that and it's not because of my code. I fixed one minor thing related to that here: https://github.com/x-tools/xtools/pull/143/files#diff-46cfccde4e078af57b974da894727b03R90

Proof that master produces same error:

image

Albert221:
Everything fixed. Added `deleted` parameter to API with corresponding docs.

But please tell me, wouldn't adding these two API calls break backwards compatibility of the API? If someone will want to pass `order` parameter using the old URL, it will pass to `deleted` which will end up in always returning the first portion of `both` deleted and live pages. Or we don't care about this simple API BC?

MusikAnimal:
> wouldn't adding these two API calls break backwards compatibility of the API?

Yep haha. The old route for the actual Pages Created tool will also be semi-broken, but we only just introduced the `offset` parameter maybe a week ago. Hopefully there aren't very many links that will be broken, but I'd rather break them then put "offset" before "deleted". I suppose this is mainly an issue with our routing system. Oh well!

> It happens on master too, I double checked that and it's not because of my code. I fixed one minor thing related to that

Yeah that's why I was so confused! I couldn't figure out what in your code was causing it. Thanks for partially fixing it! Don't worry about the other bug, I'll look into that.

MusikAnimal (approved):

MusikAnimal:
Works great! I shall merge :)

The only weird thing is that we can't tell if deleted pages are redirects. So if "deleted only" is selected, the redirects option should be set to "exclude redirects" and disabled. I don't think that's important for this PR... it'd require more backend work (in the controller, if `deleted` is passed in, it should redirect to `.../noredirects/deleted`), and some JavaScript on the form. You are welcome to make a follow-up PR for that, but it is not necessary :) We'll fix it later.

I'd like to thank you for helping out with XTools. This was my first experience with the GCI program and I very much enjoyed it. I hope you did too!

I will be back from holiday on January 6. I think that leaves a week or two of GCI, and I'd be happy to add more XTools tasks. In the meantime Sam or Matthew may be willing to add more, but I will not speak for them. If you'd like to work on XTools outside of GCI, that's of course okay too! :) All the tasks under the "Ready" column are up for grabs: https://phabricator.wikimedia.org/project/view/803/

PR #149, opened 2018-01-10
Add setting dates range for Page History

Albert221:
Bug: T181694


src/AppBundle/Controller/ArticleInfoController.php
```diff
@@ -121,13 +121,22 @@ public function gadgetAction(Request $request)
 
     /**
      * Display the results.
-     * @Route("/articleinfo/{project}/{article}", name="ArticleInfoResult", requirements={"article"=".+"})
+     * @Route(
+     *     "/articleinfo/{project}/{start}/{end}/{article}", name="ArticleInfoResult",
```
MusikAnimal: This routing change will be a problem :( ArticleInfo is very popular and there are a lot of links to it. As optional parameters it'd be better if the start/end dates are at the end, but I suppose there's the caveat that "My article/2017-01-23/2018-01-23" is a valid page title.

Maybe we should use query string parameters instead?

src/Xtools/ArticleInfo.php
```diff
@@ -21,6 +21,12 @@ class ArticleInfo extends Model
     /** @var Page The page. */
     protected $page;
 
+    /** @var false|int From what date to obtain records. */
+    protected $startDate;
+
+    /** @var false|string To what date to obtain records. */
```
MusikAnimal: `string` or `int`?

src/Xtools/ArticleInfo.php
```diff
@@ -334,7 +362,59 @@ public function topTenPercentage()
      */
     public function getPageviews($latest)
     {
-        return $this->page->getLastPageviews($latest);
+        if (false === $this->startDate && false === $this->endDate) {
+            return $this->page->getLastPageviews($latest);
+        }
+
+        list($start, $end) = $this->translateDatesToYYYYMMDD($this->startDate, $this->endDate);
+        list($start, $end) = $this->applyDatesDefaults($start, $end, $latest);
```
MusikAnimal: I'm confused why we're using `$latest` in this case. The pageviews should be for the same date as what was requested. So if they only gave an end date, it'd use the date of the first revision to the specified end date.

src/Xtools/ArticleInfoRepository.php
```diff
@@ -36,21 +41,51 @@ public function getBotData(Page $page)
...
+    /**
+     * Creates WHERE conditions with date range to be put in query.
+     *
+     * @param false|int $start
+     * @param false|int $end
+     * @param string $tableAlias Alias of table FOLLOWED BY DOT.
+     * @param string $field
+     * @return string
+     */
+    public static function createDatesConditions($start, $end, $tableAlias = '', $field = 'rev_timestamp')
```
MusikAnimal: What if we moved this to `Repository`? Then it's accessible in PageRepository, ArticleInfoRepository, etc., and it doesn't have to be static.

MusikAnimal (changes requested):
Looks awesome! I left a few comments. This is not a complete review. I unfortunately might be away much of the day, but I'll definitely get to this as soon as possible.

src/AppBundle/Controller/ArticleInfoController.php
MusikAnimal: Ideally we could use some trickery to handle the "My article/2017-01-23/2018-01-23" edge case, or perhaps we don't care... Interested to hear your thoughts.

src/Xtools/ArticleInfo.php
Albert221: That was what I meant when I told you on IRC that I have minor logical-thinking problems :D

src/Xtools/ArticleInfoRepository.php
Albert221: I second that

MusikAnimal:
Now that I'm finally home, I'm able to do a more thorough review. A few more issues, I'm afraid:
* "Page size" is not that of the last edit. In the view, instead of calling `page.length` I would do `ai.length`, and that method in ArticleInfo will conditionally call `$this->page->getLength()` or `$this->lastEdit->getLength()` based on whether a date range was provided. Probably should have some sort of `hasDateRange()` method in ArticleInfo, too.
* "Total edits" does not respect the date range. For this I suppose we need to keep a running total of the number of edits in the `while` loop of `ArticleInfo::parseHistory()`, setting it on `$this->numRevisionsProcessed`, and then hopefully `getNumRevisions()` will just work (because it is called after `ArticleInfo::prepareData()` is called).
* The "Month counts" section always goes through December, even if you put say, July, as the end date.
* Super duper minor priority -- the Pageviews links at the top ("Pageviews (All languages · Redirects)") should pass in the date range to the Pageviews tool. This would mean allowing for start/end parameters in the `wiki.pageviewsLinks` macro. Definitely don't worry about this! I can do it :)

Some things simply need to be hidden when there is a date range:
* "Assessment" and "Bugs" -- both in General stats and the dedicated `<section>`'s.
* "Page watchers"
* The "(60 days)" part of the "Pageviews (60 days)" label.
* Possibly the "Edits in the past..." fields? The numbers are right, but are relative to the end date, so the language "in the past" might be confusing. I'm definitely OK just hiding them altogether, as I don't think that information is as relevant when working within a date range.

If you need/want help with ANY of this please feel free to ask. I was worried this task was a bit too big, so don't feel like you have to do everything!

Albert221:
I localized issue of having all months to be here: https://github.com/x-tools/xtools/blob/d0838b9a2a81f52c90424d970d9b316d057d50b2/src/Xtools/ArticleInfo.php#L851-L867

web/static/css/application.scss
```diff
@@ -218,6 +218,11 @@ $tooltip-shadow: #ccc;
         text-decoration: none;
     }
 }
+.date-range {
+    @include center('vertical');
+    right: 0;
+    font-size: 70%;
+}
```
MusikAnimal: Minor complaint: We could use some mobile-friendliness here, something like:
```css
@media (max-width: $bootstrap-sm) {
    .date-range {
        display: block;
        position: relative;
        top: 0;
        transform: none;
    }
}
```

app/Resources/views/articleInfo/result.html.twig
```diff
@@ -42,7 +57,7 @@
             {% if ai.assessments and not (page.isMainPage) %}
                 {% set sections = sections|merge(['assessments']) %}
             {% endif %}
-            {% if ai.numBugs > 0 %}
+            {% if ai.numBugs > 0 and not (ai.hasDateRange) %}
```
MusikAnimal: Let's do `not (ai.hasDateRange)` first, because `ai.numBugs` will run the queries.

app/Resources/views/articleInfo/result.html.twig
```diff
@@ -92,7 +107,7 @@
                         </td>
                         <td>{{ ai.numEditors|num_format }}</td>
                     </tr>
-                    {% if ai.assessments and not (page.isMainPage) %}
+                    {% if ai.assessments and not (page.isMainPage) and not (page.hasDateRange) %}
```
MusikAnimal: `ai.hasDateRange` (not `page`), and again I'd put this check before `ai.assessments`

app/Resources/views/articleInfo/result.html.twig
```diff
@@ -260,22 +277,22 @@
                         <td>{{ msg('links-external') }}</td>
                         <td>{{ ai.linksExtCount|num_format }}</td>
                     </tr>
-                    {% if page.watchers is not empty %}
+                    {% if page.watchers is not empty and not (ai.hasDateRange) %}
```
MusikAnimal: `not (ai.hasDateRange)` first. Here I think `Page::getPageInfo()` has already been ran, but good to be consistent with the placement of the hasDateRange check.

app/Resources/views/articleInfo/result.html.twig
```diff
@@ -260,22 +277,22 @@
...
-                        <td>{{ msg('pageviews') }} (60 {{ msg('num-days', [60]) }})</td>
+                        <td>{{ msg('pageviews') }}{% if not (ai.hasDateRange) %} (60 {{ msg('num-days', [60]) }}){% endif %}</td>
                         <td>
                             {{ wiki.pageviewsLink(page, ai.pageviews(pageviewsOffset)|num_format, 'range=latest-' ~ pageviewsOffset) }}
                         </td>
                     </tr>
                     {% endif %}
 
-                    {% if ai.numBugs > 0 %}
+                    {% if ai.numBugs > 0 and not (ai.hasDateRange) %}
```
MusikAnimal: `not (ai.hasDateRange)` first

app/Resources/views/articleInfo/result.html.twig
```diff
@@ -753,7 +770,7 @@
         {% endif %}
 
         {# ======================== ASSESSMENTS ======================== #}
-        {% if ai.assessments and not (page.isMainPage) %}
+        {% if ai.assessments and not (page.isMainPage) and not (ai.hasDateRange) %}
```
MusikAnimal: `not (ai.hasDateRange)` first

app/Resources/views/articleInfo/result.html.twig
```diff
@@ -792,7 +809,7 @@
         {% endif %}
 
         {# ======================== BUGS ======================== #}
-        {% if ai.numBugs > 0 %}
+        {% if ai.numBugs > 0 and not (ai.hasDateRange) %}
```
MusikAnimal: `not (ai.hasDateRange)` first

app/Resources/views/macros/wiki.html.twig
```diff
@@ -217,17 +217,17 @@
...
-        msg('pageviews')
+        msg('pageviews') ~ start is not null ? ('&start=' ~ start) : '' ~ end is not null ? ('&end=' ~ end) : ''
     )}} ({{ extLink(
         'https://tools.wmflabs.org/langviews?project=' ~ page.project.domain ~ '&page=' ~ page.title,
-        msg('all-languages')
+        msg('all-languages') ~ start is not null ? ('&start=' ~ start) : '' ~ end is not null ? ('&end=' ~ end) : ''
     )}} &middot; {{ extLink(
         'https://tools.wmflabs.org/redirectviews?project=' ~ page.project.domain ~ '&page=' ~ page.title,
-        msg('redirects')
+        msg('redirects') ~ start is not null ? ('&start=' ~ start) : '' ~ end is not null ? ('&end=' ~ end) : ''
```
MusikAnimal: Looks like the `~ start ...` is next to the link text and not in the URL.

app/Resources/views/macros/wiki.html.twig
```diff
@@ -217,17 +217,17 @@
     {{ extLink(url, msg('pageviews'), label) }}
 {% endspaceless %}{% endmacro %}
 
-{% macro pageviewsLinks(page) %}{% spaceless %}
+{% macro pageviewsLinks(page, start, end) %}{% spaceless %}
```
MusikAnimal: You might consider setting the defaults `start = null, end = null`, then we don't need the null checks at https://github.com/x-tools/xtools/pull/149/files#diff-fefe304fbb1893557366c352e22673cfR42

src/AppBundle/Controller/ArticleInfoController.php
```diff
@@ -120,15 +120,36 @@ public function gadgetAction(Request $request)
...
+        // for router regex parser to parse `article` with those parameters and then
+        // manually retrieve what we want. It's done this way because programmatical way
+        // is much easier (or maybe even only existing) solution for that.
+
+        // Does path have `start` and `end` parameters (even empty ones)?
+        if (1 === preg_match('/(.+?)\/(|\d{4}-\d{2}-\d{2})\/(|\d{4}-\d{2}-\d{2})$/', $article, $matches)) {
+            $article = $matches[1];
+            $start = $matches[2];
+            $end = $matches[3];
+        }
```
MusikAnimal, line 151: :100:

app/Resources/views/articleInfo/result.html.twig
MusikAnimal: Also need the `not (ai.hasDateRange)` check on line 57

MusikAnimal (changes requested):
Looking really good!! Just a few more minor complaints :) Thanks so much, I know this is a lot of work!

MusikAnimal (approved):

PR #150, opened 2018-01-11
Add minus-x to composer.json

Albert221:
Bug: T184600


MusikAnimal:
Good stuff! `composer test` is actually ran manually and not in the CI build, because -- no apparent reason. I'm going to merge this as-is then clean up the Scrutinizer and Travis configs.

PR #181, opened 2018-03-13
Add Korean texts to (semi-)auto edit list

revi:
I am regex newbie (despite editing regex for few years) so review needed for there, despite I think you aren't necessarily good at Korean. Or I can find someone who does Korean and regex.

### DIFF
--- a/app/config/semi_automated.yml
+++ b/app/config/semi_automated.yml
@@ -94,11 +94,9 @@ parameters:
     ko:
         Undo:
             regex: '판 편집을 되돌림'
-            tag: mw-undo
             label: '편집 취소'
         Generic rollback:
             regex: '(의 마지막 판으로 되돌림|의 편집을 전부 되돌림)'
-            tag: mw-rollback
             label: '되돌리기'
         Page move:
             regex: '문서로 이동했습니다'
@@ -357,15 +355,15 @@ parameters:
             link: User:Evad37/MoveToDraft
     ko.wikipedia.org:
         AutoWikiBrowser:
-            regex: 'using [[위키백과:AWB|AWB]]'
+            regex: 'using \[\[위키백과:AWB\|AWB\]\]'
             link: 위키백과:AWB
             label: 오토위키브라우저
         Popups:
-            regex: '[[백:팝업|NavPop]]'
+            regex: '\[\[백:팝업\|NavPop\]\]'
             link: 백:팝업
             label: 내비게이션 팝업
         Twinkle:
-            regex: '[[위키백과:도구/트윙클|TW]]'
+            regex: '\[\[위키백과:도구\/트윙클\|TW\]\]'
             link: 위키백과:도구/트윙클
             label: 트윙클
         WikiLove:
--- a/app/config/semi_automated.yml
+++ b/app/config/semi_automated.yml
@@ -91,6 +91,18 @@ parameters:
         Redirect:
             regex: 'Weiterleitung nach \[\[.*?erstellt'
             label: Weiterleitung
+    ko:
+        Undo:
+            regex: '판 편집을 되돌림'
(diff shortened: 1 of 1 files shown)


app/config/semi_automated.yml
```diff
@@ -91,6 +91,18 @@ parameters:
         Redirect:
             regex: 'Weiterleitung nach \[\[.*?erstellt'
             label: Weiterleitung
+    ko:
+        Undo:
+            regex: '판 편집을 되돌림'
+            tag: mw-undo
```
MusikAnimal: This gets merged into the global rule, so the tag specification isn't needed here, same with rollback below

app/config/semi_automated.yml
```diff
@@ -343,6 +355,23 @@ parameters:
         MoveToDraft:
             regex: '\[\[User:Evad37\/MoveToDraft\.js\|via script'
             link: User:Evad37/MoveToDraft
+    ko.wikipedia.org:
+        AutoWikiBrowser:
+            regex: 'using [[위키백과:AWB|AWB]]'
```
MusikAnimal: Need to escape the brackets and pipes, so `'using \[\[위키백과:AWB\|AWB\]\]'`

app/config/semi_automated.yml
```diff
@@ -343,6 +355,23 @@ parameters:
...
+            link: 위키백과:AWB
+            label: 오토위키브라우저
+        Popups:
+            regex: '[[백:팝업|NavPop]]'
```
MusikAnimal: `'\[\[백:팝업\|NavPop\]\]'`

app/config/semi_automated.yml
```diff
@@ -343,6 +355,23 @@ parameters:
...
+            link: 백:팝업
+            label: 내비게이션 팝업
+        Twinkle:
+            regex: '[[위키백과:도구/트윙클|TW]]'
```
MusikAnimal: `'\[\[위키백과:도구\/트윙클\|TW\]\]'`

app/config/semi_automated.yml
```diff
@@ -343,6 +355,23 @@ parameters:
...
+            link: 위키백과:도구/트윙클
+            label: 트윙클
+        WikiLove:
+            regex: '새 위키사랑 글'
+            tag: 'wikilove'
```
MusikAnimal, line 371: I guess this tag should be moved to the global ruleset, but I'll take care of that.

We also need a page to link to. If there is no page, I guess let's link to `mw:Extension:WikiLove/ko`

MusikAnimal (changes requested):

app/config/semi_automated.yml
revi: tag: isn’t correctly catching old (pre-tag era) edit summaries so I thought I would put that there.

MusikAnimal: Correct, but any per-wiki or per-language rules get merged into global ones, e.g. https://github.com/x-tools/xtools/pull/181/files#diff-03baa01bde8a3fd8340f8f6d32967575R34

Adding it here won't hurt anything, though :) The merging thing is just a means to keep the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

revi: OK, so I think `tag:` is the only thing that needs to be removed?

app/config/semi_automated.yml
revi: For Korean Wikipedia, I think `ko:위키백과:위키사랑` is short but better than almost-not-translated `mw:Extension:WikiLove/ko`.

MusikAnimal:
There are a few other tiny issues, but I can take it from here. Many thanks!

PR #184, opened 2018-03-20
Fix for a negative day on AdminScore.

Matthewrbowker:
Bug: T190201


MusikAnimal:
I honestly don't know how these formulas are supposed to work, so I'll take your word that this fixes it :)

FYI I squashed the commits here in the GitHub interface before merging.

PR #202, opened 2018-06-23
Fix for control characters in database results

Matthewrbowker:
Bug: T197165


Matthewrbowker:
Please squash when merging... I had a git-ism and it looks pretty weird xD

src/AppBundle/Controller/AdminScoreController.php
```diff
@@ -177,6 +177,8 @@ public function resultAction(Request $request)
             $key = $row['source'];
             $value = $row['value'];
 
+            $value = str_replace("\x00", "", $value);
```
MusikAnimal: Let's add a comment with the task number so there's some context for why we're doing this.

MusikAnimal:
Merged :) However I'm not convinced this is an issue with the database, and I see the DBAs already removed the DBA label from the task. Whatever the problem is, it probably happens elsewhere and not just for AdminScore?

Matthewrbowker:
I am confused, because there is nowhere else the problem could exist except within the database.  In my memory (which I will verify), AdminScore is the only place that feeds direct database values into Twig without doing anything first.  Everywhere else pulls the database values and does something with them.

PR #203, opened 2018-06-25
Allow the wiki table to be configured by parameter

Matthewrbowker:
See T178343.


Matthewrbowker:
This will need to be squished when merged.

Sam and Leon, I added you both because this is my first foray into the repositories.

src/Xtools/ProjectRepository.php
```diff
@@ -115,9 +115,15 @@ public function getAll()
             return $this->cache->getItem($this->cacheKeyAllProjects)->get();
         }
 
+        if ($this->container->hasParameter("database_meta_table")) {
+            $table = $this->container->getParameter("database_meta_table");
+        } else {
+            $table = "wiki";
+        }
+
```
MusikAnimal, line 123: Should we not put this in `table_map.yml` and check that when pulling from `'wiki'`? I guess that's a little weird, because we're currently only using the table mapping for the "normal" databases containing revisions and logging.

Otherwise I would suggest moving the logic to a private method in this class (say `getMetaTable`) so that it can be reused, but not a big deal :)

Matthewrbowker: Fair enough, may move the logic.  In making this pull request, I was attempting to replicate the existing functionality with the added option of the new table.  The existing functionality had "wiki" hard-coded, so that's what I went with.  

As for table mapping, I figured it would be easier to explain and document if we went with "database_meta_name" and "database_meta_table" being related values defined in the same place, as opposed to obfuscating the information behind mapping.

MusikAnimal: Yeah you're right. The new meta-related parameter make sense to be alongside the existing one. No need to refactor that logic if you don't want to, just a suggestion in case we wanted to reference the wiki table elsewhere. The current code doesn't break the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), after all :)

MusikAnimal:
Exciting to see us one step closer to supporting Miraheze! There are probably a lot of places I forgot to put `is_labs` checks, so let me know if things are broken and you need help :)

PR #223, opened 2018-09-27
Change default/max range in AdminStats from 30 to 31 days

Amorymeltzer:
Some of the best months have 31 days


MusikAnimal:
Looks great, thank you! It may be a bit before this gets deployed (week maybe), since we're in the middle of moving over to new servers and upgrading to Symfony 4.

PR #232, opened 2018-11-23
Improve pt translation and fix some links

he7d3r:


he7d3r:
@MusikAnimal : could you review this?

MusikAnimal:
Thanks for the PR! Everything looks good code-wise, but I still need to test. There are a few other requests for changes to this config file. I'll go through them all probably today or tomorrow.

MusikAnimal (approved):
Sorry for taking so long to get to this! Looks great. I couldn't find an example of where reFill was used on ptwiki, but all you changed here was the link, and that should work. Thanks for contributing! :D

PR #241, opened 2019-01-14
ArticleInfo: Fix prod error thrown when checking Main Page

MusikAnimal:
Rework some CSS rules to have better cross-browser support


app/Resources/assets/css/application.scss
```diff
@@ -246,7 +246,7 @@ $tooltip-shadow: #ccc;
     }
 
     .bold {
-        font-weight: 500;
+        font-weight: bold;
```
samwilson, line 249: Is `bolder` better here? It is relative to the context.

MusikAnimal: Yay, Sam is back on XTools! :)

I don't think I even knew about `bolder`. I don't even know where `500` came from... I just noticed after moving to Linux, it has no effect in Chromium. `bolder` does work, but it's not any bolder than `bold`. Anyway, here I think all we want is normal boldness. That was the intention, I believe.

MusikAnimal:
I only created this PR to make sure tests passed, but I'm thrilled to see some participation from others! I'm going to go ahead and merge assuming it's OK.

I've got a few more PRs coming to fix up all the prod errors we've had over the past few weeks.

PR #242, opened 2019-01-19
[AutoEdits] Update AFC/R HS

JJMC89:
Add User:Enterprisey/AFCRHS to AFC/R HS

Bug: https://phabricator.wikimedia.org/T214137


MusikAnimal:
Thanks! FYI, for XTools I use `Bug: T123` because the repo is mirrored on Diffusion, so commits will show up in the relevant task if you use T123 instead of the full URL. Not important, just letting you know! :)

PR #244, opened 2019-01-19
Add another regex of Sagittarius (a fork)

DannyS712:
I also changed the link to the page of the fork's maintainer (Sam Sailor), since the original script's maintainer is inactive. This is my first attempt to use github, so if I did something wrong I'm sorry.


DannyS712:
See phabricator ticket: https://phabricator.wikimedia.org/T212984

PR #248, opened 2019-01-22
autoedits - add another script installer

DannyS712:
Add a regex for https://en.wikipedia.org/wiki/User:Enterprisey/script-installer to the regex for Equazcion's script installer, change the link from Equazcion's page to Enterprisey's because Equazcion hasn't edited in over 2 years. This solves https://phabricator.wikimedia.org/T212927


MusikAnimal:
Closing per https://github.com/x-tools/xtools/pull/267#issuecomment-484334159

PR #263, opened 2019-04-06
Show live revisions count instead of total revisions

QEDK:
Tiny fix

### DIFF
--- a/templates/editCounter/general_stats.html.twig
+++ b/templates/editCounter/general_stats.html.twig
@@ -474,7 +474,7 @@
             chart.pie_chart('deleted_edits',
                 [{
                     label: msg('live-edits'),
-                    value: ec.countAllRevisions,
+                    value: ec.countLiveRevisions,
                     percentage: ((ec.countLiveRevisions / ec.countAllRevisions) * 100)
                 },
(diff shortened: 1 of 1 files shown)


templates/editCounter/general_stats.wikitext.twig
```diff
@@ -30,7 +30,7 @@
 {% endif %}
 |-
 | {{ msg('is-admin') }}
-| {% if user.isAdmin(project) %}1{% else %}0{% endif %}
+| {% if user.isAdmin(project) %}Yes{% else %}No{% endif %}
```
MusikAnimal: This needs to be localized. I can show you how to do this if you'd like, but regardless I think this change should be saved for a separate PR. We could even show images for a green tick and red cross, like we do in the UI (but putting "yes" / "no" as the `alt` text, for accessibility).

MusikAnimal (changes requested):
Thank you for the fix!

If we can move the yes/no change to a separate PR then I will merge :)

QEDK:
@MusikAnimal done.

MusikAnimal:
Thank you!

PR #253, opened 2019-02-25
+CatAlot

DannyS712:
Add another semi-automated tool, CatAlot; solves phab https://phabricator.wikimedia.org/T217091

PR #251, opened 2019-01-25
autoedits: add effp-helper to enwiki

DannyS712:
Add suffusion of yellow's edit filter false positives helper, solves https://phabricator.wikimedia.org/T214005

PR #250, opened 2019-01-25
autoedits: add reply link to enwiki

DannyS712:
Solves https://phabricator.wikimedia.org/T212925

PR #249, opened 2019-01-23
Autoedits: Add short description helper

DannyS712:
Solves https://phabricator.wikimedia.org/T213019
In one of the intermediate commits I added a different tool (wikiplus), but later removed it (so please ignore)

PR #267, opened 2019-04-15
Autoedits: multiple additions and updates

DannyS712:
Add effp-helper, reply link, and short description helper. Update regex for CatALot, add another regex for script installer, and change the link to Enterprisey's version. Replaces #253, #251, #250, #249, and #248 (combines the commits, corrects for merge conflict, adds namespace specifications).

Bugs: T217091, T214005, T212925, T213019, and T212927.

### DIFF
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -455,7 +455,8 @@ parameters:
         Reply link:
             regex: '\(\[\[w:en:User:Enterprisey\/reply-link\|reply-link\]\]\)'
             link: User:Enterprisey/reply-link
-            talk_namespaces: true
+            talk_namespaces: true,
+            contribs: true
         Short description:
             regex: '\(\[\[User:Galobtter\/Shortdesc helper\|Shortdesc helper\]\]\)'
             link: User:Galobtter/Shortdesc helper
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -67,7 +67,7 @@ parameters:
             regex: '\|HotCat\]\]|Gadget-Hotcat(?:check)?\.js\|Script|\]\] via HotCat|\[\[WP:HC\|'
             link: Special:MyLanguage/Project:HotCat
         Cat-a-lot:
-            regex: '\[\[Help:Cat-a-lot\|Cat-a-lot\]\]'
+            regex: '\[\[Help:Cat-a-lot\|Cat-a-lot\]\]|\(CatAlot\)|\[\[c:Help:Cat-a-lot\|Cat-a-lot\]\]'
             link: commons:Cat-a-lot
         Global rename:
             regex: 'Automatically moved page while renaming the user'
@@ -354,8 +354,8 @@ parameters:
             regex: '\[\[w:en:WP:RLR\|You can help!'
             link: en:WP:RLR
         Script Installer:
-            regex: '\[\[User:Equazcion\/ScriptInstaller\|Script Installer'
-            link: User:Equazcion/ScriptInstaller
+            regex: '\[\[User:Equazcion\/ScriptInstaller\|Script Installer|\(\[\[User:Enterprisey\/script-installer\|script-installer\]\]\)'
+            link: User:Enterprisey/script-installer
             namespaces: [2]
         findargdups:
             regex: '\[\[:en:User:Frietjes\/findargdups'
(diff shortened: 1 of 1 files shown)


config/semi_automated.yml
```diff
@@ -449,6 +449,18 @@ parameters:
             tag: 'OAuth CID: 860'
             link: User:Enterprisey/YABBR
             namespaces: [0]
+        effp-helper:
+            regex: '\(\[\[User:Suffusion of Yellow\/effp-helper(\.js)?\|effp-helper\]\]\)'
+            link: User:Suffusion of Yellow/effp-helper
+            namespaces: [0]
+      Reply link:
```
MusikAnimal: Need to indent two more spaces here and on line 460

config/semi_automated.yml
```diff
@@ -449,6 +449,18 @@ parameters:
             tag: 'OAuth CID: 860'
             link: User:Enterprisey/YABBR
             namespaces: [0]
+        effp-helper:
+            regex: '\(\[\[User:Suffusion of Yellow\/effp-helper(\.js)?\|effp-helper\]\]\)'
+            link: User:Suffusion of Yellow/effp-helper
+            namespaces: [0]
```
MusikAnimal: This could be used in any namespace, right?

config/semi_automated.yml
```diff
@@ -67,7 +67,7 @@ parameters:
             regex: '\|HotCat\]\]|Gadget-Hotcat(?:check)?\.js\|Script|\]\] via HotCat|\[\[WP:HC\|'
             link: Special:MyLanguage/Project:HotCat
         Cat-a-lot:
-            regex: '\[\[Help:Cat-a-lot\|Cat-a-lot\]\]'
+            regex: '\[\[Help:Cat-a-lot\|Cat-a-lot\]\]|\(CatAlot\)|\[\[c:Help:Cat-a-lot\|Cat-a-lot\]\]'
```
MusikAnimal: I think we can get away with just
```
Help:Cat-a-lot\|Cat-a-lot\]\]|\(CatAlot\)
```
The basic idea is to make the regex as short yet simple as possible, while preventing false positives. Here it seems unlikely someone would manually put `Help:Cat-a-lot|Cat-a-lot]]` in their edit summary.

MusikAnimal (changes requested):
Thanks for this! I've left a few comments. I'm still reluctant to add more tools, because AutoEdits is crazy slow... but I've accepted there's nothing we can do about this until https://phabricator.wikimedia.org/T215445 is resolved.

Did you want to abandon your other PRs that this replaces? #253, #251, #250, #249 and #248

config/semi_automated.yml
DannyS712: Done

config/semi_automated.yml
DannyS712: Yes. I originally had it in ns:0 only since that is likely where its used most, but I have removed that limitation

config/semi_automated.yml
DannyS712: done

DannyS712:
@MusikAnimal all done. Yes, I would abandon #248, #249, #250, #251, and #253. I figured it would be easier to do them all at once, and anyway those branches were out of date with the namespace configurations.

config/semi_automated.yml
```diff
@@ -449,6 +449,17 @@ parameters:
             tag: 'OAuth CID: 860'
             link: User:Enterprisey/YABBR
             namespaces: [0]
+        effp-helper:
+            regex: '\(\[\[User:Suffusion of Yellow\/effp-helper(\.js)?\|effp-helper\]\]\)'
+            link: User:Suffusion of Yellow/effp-helper
+        Reply link:
+            regex: '\(\[\[w:en:User:Enterprisey\/reply-link\|reply-link\]\]\)'
+            link: User:Enterprisey/reply-link
+            talk_namespaces: true
```
MusikAnimal: I forgot one thing... I think we should add `contribs: true` for Reply link. This will make it still show the edits in the non-automated contributions list. AutoEdits is usually used to filter out the noise of semi-automated edits, and only see relevant substantial contributions. An example is ProveIt; while it is a semi-automated tool, you can still make large content changes which users may wish to see. Similarly users might want to see comments made using Reply link. Hopefully that makes sense... :)

MusikAnimal (changes requested):
Sorry, one last thing!

config/semi_automated.yml
DannyS712: @MusikAnimal done

MusikAnimal (approved):

MusikAnimal:
Thank you, as always!

MusikAnimal:
@DannyS712 One other tiny note, this repo is [mirrored on Diffusion](https://phabricator.wikimedia.org/source/tool-xtools/), so if you put `Bug: T123` (where 123 is the task number) the commits will show up on the task. If there are multiple tasks as there are here, just chain them line by line. I did this for you when I merged your PR.

I recently made https://github.com/x-tools/xtools/blob/master/CONTRIBUTING.md to explain this.

PR #269, opened 2019-04-17
ArticleInfo: Updated icons

pythoncoder42:
* Changed Featured icon to be Norro style 1 like the rest
* Changed deprecated B+ class to have same color/icon as B-class
* Set Future, Current, SetIndex, and NA to be their actual respective icons

### DIFF
--- a/config/assessments.yml
+++ b/config/assessments.yml
@@ -132,7 +132,7 @@ parameters:
       wikiproject_prefix: 'Wikipedia:WikiProject '
       class:
         FA:
-          badge: 'b/bc/Featured_article_star.svg'
+          badge: '2/23/Symbol_star_FA.svg'
           color: '#6699ff'
           category: 'Category:FA-Class articles'
         A:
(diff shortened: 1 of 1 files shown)


pythoncoder42:
Failed test was an assert regarding a string I changed

config/assessments.yml
```diff
@@ -132,7 +132,7 @@ parameters:
       wikiproject_prefix: 'Wikipedia:WikiProject '
       class:
         FA:
-          badge: 'b/bc/Featured_article_star.svg'
+          badge: '2/23/Symbol_star_FA.svg'
```
MusikAnimal, line 135: This doesn't seem to be the standard icon used on the English Wikipedia. The icons were gathered from https://en.wikipedia.org/wiki/Wikipedia:Content_assessment, and these are the ones you see on talk pages, supplied by https://en.wikipedia.org/wiki/Template:WPBannerMeta. The classes below GA of course don't have official icons, so I just found suitable ones.

MusikAnimal (changes requested):
Thanks for this! I left a comment about the FA icon, everything else looks terrific.

> Failed test was an assert regarding a string I changed

The relevant line is https://github.com/x-tools/xtools/blob/master/tests/AppBundle/Model/PageAssessmentsTest.php#L75 if you want to update it, but per above I'm not sure about the change to the FA icon. Either way, for future reference, I can handle any test failures for you, if you're not comfortable fixing them :)

MusikAnimal:
I'm going to merge and just change the FA icon back to it's original one, per above. Thanks for the PR! :)

PR #270, opened 2019-04-17
Base: Use date_format instead of num_format for (c) years in footer

rchard2scout:
num_format causes thousands separators to be inserted (so the footer says
"© 2,008-2,019"). 2008 and 2019 are dates, so date_format makes sense and is
semantically correct.

Just using "2008" will cause PHP to interpret the string as an integer, which
gives incorrect output, so "jan-2008" is used instead. This doesn't cause a
visible change since only the year is displayed.

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -184,7 +184,7 @@
         <div class="footer-about">
             <span>
                 {# The release process is documented at https://xtools.readthedocs.io/en/latest/development.html#releases #}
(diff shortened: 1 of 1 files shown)


MusikAnimal (approved):
Nice catch, and clever solution! `num_format` was added this past February in an attempt to better localize numerals (such as for Hindu-Arabic), and I obviously overlooked what it did to westernized Arabic (such as for English). `date_format` accomplishes the same and indeed won't incorrectly add thousand separators. Thank you! :)

PR #274, opened 2019-04-30
Autoedits: Add undo-last-edit

DannyS712:
No restrictions on namespace
Bug: T222134

### DIFF
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -461,7 +461,7 @@ parameters:
             regex: '\(\[\[User:Galobtter\/Shortdesc helper\|Shortdesc helper\]\]\)'
             link: User:Galobtter/Shortdesc helper
             namespaces: [0]
-        Undo-last-edit:
+        undo-last-edit:
             regex; '\(\[\[User:Enterprisey\/undo-last-edit\.js\|undo-last-edit\]\]\)'
             link: User:Enterprisey/undo-last-edit
     ko.wikipedia.org:
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -462,8 +462,9 @@ parameters:
             link: User:Galobtter/Shortdesc helper
             namespaces: [0]
         undo-last-edit:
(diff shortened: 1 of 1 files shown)


config/semi_automated.yml
```diff
@@ -461,6 +461,9 @@ parameters:
             regex: '\(\[\[User:Galobtter\/Shortdesc helper\|Shortdesc helper\]\]\)'
             link: User:Galobtter/Shortdesc helper
             namespaces: [0]
+        Undo-last-edit:
```
MusikAnimal: should be lowercased, according to https://en.wikipedia.org/wiki/User:Enterprisey/undo-last-edit

config/semi_automated.yml
```diff
@@ -461,6 +461,9 @@ parameters:
...
+            regex; '\(\[\[User:Enterprisey\/undo-last-edit\.js\|undo-last-edit\]\]\)'
+            link: User:Enterprisey/undo-last-edit
```
MusikAnimal, line 466: This is missing a label. AutoEdits should probably fall back to using the key for the tool (`undo-last-edit`) if a label is missing, but for now it needs to be added explicitly. So all we need is `label: undo-last-edit`

MusikAnimal (changes requested):

config/semi_automated.yml
DannyS712: Line 12 says that "label: Text to use as the label, defaults to the tool's key. Use this for translations." and most of the other tools don't have labels - is this needed?

config/semi_automated.yml
DannyS712: Done

config/semi_automated.yml
MusikAnimal: @DannyS712 Bah you're right. I was going off of the test failure. Turns out the culprit is `regex;`, which should be `regex:` :)

Also, since this is revert-only, let's add `revert: true`

MusikAnimal (changes requested):

config/semi_automated.yml
DannyS712: Done

PR #276, opened 2019-05-02
AudoEdits: optimize namespaces

DannyS712:
Set namespaces for WikiLove, delsort, editProtectedHelper, For the Common Good, and IABot. Remove current specification from XFDC, and remove the wikipedia namespace from the specification for stubsearch.

Bug: T222323


MusikAnimal:
This is fantastic! Everything looks good. Thank you :)

PR #277, opened 2019-05-02
AutoEdits: General cleanup of file

DannyS712:
Alphabetize tools lists, separate global overrides from local tools, move commons out of the wikipedias and group it with wikidata, etc. No change in functionality should have occurred.
Bug: T222327

### DIFF
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -115,9 +115,8 @@ parameters:
         Undo:
             regex: 'الرجوع عن التعديل \d+ بواسطة'
     de:
-        Undo:
-            regex: 'Änderung \d+ von \[\[.*?rückgängig gemacht'
-            label: Rückgängig
+        AutoWikiBrowser:
+            regex: 'mit \[\[Project:AWB\|AWB'
         Generic rollback:
             regex: 'Änderungen von \[\[.*?die letzte Version von \[\[.*?zurückgesetzt'
             label: Zurücksetzen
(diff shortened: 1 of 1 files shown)


DannyS712:
@MusikAnimal I can't tell why this fails - every site I asked says that the YAML is valid...

config/semi_automated.yml
```diff
@@ -161,314 +169,265 @@ parameters:
...
-            regex: '\[\[MediaWiki:Gadget-SettingsManager.js\|SettingsManager'
-            link: MediaWiki:Gadget-SettingsManager.js
-        RotateLink:
-            regex: '\(\[\[Help:RotateLink\|Script'
-            link: Help:RotateLink
-    de.wikipedia.org:
+        WPCleaner:
+            link: ويكيبيديا:وب_كلينر
+            label: وب_كلينر
+        de.wikipedia.org:
```
MusikAnimal: `de.wikipedia.org:` is indented one tab too many. This is why the test was failing :)

MusikAnimal (changes requested):
> MusikAnimal I can't tell why this fails - every site I asked says that the YAML is valid...-

See the comment. I guess the other YAML validators weren't checking for duplicate keys! Because that would not be valid.

config/semi_automated.yml
DannyS712: Done, but that doesn't appear to have been the only issue...

MusikAnimal: This time it was a fluke with Scrutinizer. I triggered another build and it passed.

I'll try to get to doing a full review soon.

MusikAnimal (approved):
This was very tedious to review, but I manually went through everything and it looks good. Thank you for making separate commits, that helped (though I will still squash them when merging).

PR #279, opened 2019-05-05
i18n: clarify pc-reject

DannyS712:
Change pc-reject from "Pending changes reject" to "Pending changes unaccept" since the actual action that is logged is to unaccept the revision. For example, Patroller stats currently lists DannyS712 as having 0 "Pending changes reject", while autoedits lists the same user as having 153 "Pending changes revert edits". This is because autoedits tracks the edits themselves, while patrollerstats tracks log entries. The heading for the log entry column should correspond to the actual content, in this case unaccepting revisions.


MusikAnimal:
Good call. Thanks!

PR #283, opened 2019-05-07
AutoEdits: Expand with new tools/regexes

DannyS712:
Add awb regex for arz wiki, add 2 new tools (heading editor and pageswap)
Bugs: T221520, T222679, T222680

### DIFF
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -114,6 +114,9 @@ parameters:
     ar:
         Undo:
             regex: 'الرجوع عن التعديل \d+ بواسطة'
+    arz:
+        AutoWikiBrowser:
+            regex: 'عن طريق \[\[Project:AWB\|اوب\]\]'
     de:
         AutoWikiBrowser:
             regex: 'mit \[\[Project:AWB\|AWB'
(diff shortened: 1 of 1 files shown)


DannyS712:
This keeps failing tests with "Unknown modifier 'h'" - what does that mean?

DannyS712:
@stupidsmart1 @MusikAnimal any idea when this might be merged?

MusikAnimal:
I don't think @stupidsmart1 is a real person. There's been a few bizarre GitHub bots going around lately making dubious actions in random repos.

Anyway we're in the middle of https://phabricator.wikimedia.org/T223667 right now. We need to get that deployed and see how it performs before doing any more feature work. The good news I *think* AutoEdits might actually have better performance using the new actor schema.

cyberpower678:
Yea, I was wondering who @stupidsmart1 is.  I quietly follow the updates to this repo and never saw this user before.

DannyS712:
@MusikAnimal now that T223667 has been resolved, do you think you can take a look at this?

MusikAnimal:
Sorry for the long wait! I've tested this and it works well :) I will merge this, but note that AutoEdits is still really, really slow (224 seconds for your account). I am apprehensive about adding any more tools right now. I think most of the slowness can be attributed to the [actor migration](https://wikitech.wikimedia.org/wiki/News/Actor_storage_changes_on_the_Wiki_Replicas), but the sheer length of the regex is surely part of it, too. I might be time for a spring cleaning to trim down the regular expressions. E.g. we `\(\[\[WP:DFY\|DFY\]\]\)` could probably just be `WP:DFY\|DFY`.

DannyS712:
@MusikAnimal then now would probably be a bad time to tell you that I have 2 more tools for commons, 1 for global, 1 for the galician wikipedia, and a bunch more for enwiki, right? Anyway, I've made a start at #291.

PR #291, opened 2019-06-13
AutoEdits: Start cleaning up regex

DannyS712:
Simplify regex definitions for AFCH, MoveToDraft, STiki, draftify, closetfd/closemfd, and new redirects
Per comment on #283. Hopefully this can get reviewed **before** those new tools are released in the next update.
Bug: T225691


MusikAnimal:
This is a good start. Thank you.

PR #293, opened 2019-06-20
AutoEdits: Expand Commons tools

DannyS712:
Add CropTool and update RenameLink, HotCat, and Cat-a-lot
Bugs: T226142, T226138, T226139, T226140, T226141


DannyS712:
@MusikAnimal any idea when this might be looked at?

MusikAnimal (approved):
I assume Cat-a-lot and HotCat do/should be using tags on other wikis, but this is fine for now. Sorry for taking so long to review!

PR #297, opened 2019-07-04
Blame: improve algorithm, allowing partial matches

MusikAnimal:
Don't error out on result pages if search query isn't given

Copy edit to Blame description


samwilson (commented):
I haven't run this, but it looks like it makes sense.

MusikAnimal:
Thanks for the feedback! This task really made my head hurt =p

PR #307, opened 2019-09-13
Remove pages deleted from pages section of edit counter

DannyS712:
Per reasoning on phab

Bug: T232804


MusikAnimal (changes requested):
I think this makes sense! Thanks for remembering to update the wikitext template, too.

Everything looks great except with this change we're no longer using the `pages-deleted` message. We should remove it from en.json and qqq.json (translatewiki will remove it from the other files, and you may not even need to update qqq.json).

MusikAnimal (approved):

PR #309, opened 2019-09-15
Link account creation log to newusers/create2

DannyS712:
Bug: T232963

### DIFF
--- a/templates/editCounter/general_stats.html.twig
+++ b/templates/editCounter/general_stats.html.twig
@@ -325,7 +325,7 @@
                 <tr>
                     <td>{{ msg('accounts-created') }}</td>
                     <td>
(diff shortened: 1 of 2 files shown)


MusikAnimal:
We need to link to `byemail` as well as `create2`. I'm not sure if that's possible?

For instance, User:L235 created 7 accounts: [4 as `create2`](https://en.wikipedia.org/w/index.php?title=Special:Log&action=view&user=L235&type=newusers&subtype=create2) and another [3 accounts as `byemail`](https://en.wikipedia.org/wiki/Special:Log?type=newusers&user=L235&page=&wpdate=&tagfilter=&subtype=byemail). The latter is important because it includes accounts created through ACC.

DannyS712:
I don't think that is possible currently

MusikAnimal:
> I don't think that is possible currently

With this I am going to close this PR. Thanks for putting the work into it, anyway!

PR #339, opened 2019-10-23
Make year/month count CSV exports comprehensible

MusikAnimal:
Columns are namespaces, rows are years/months.

Also removed one unused class property from EditCounter.

Bug: T236095


samwilson (approved):
Untested, but looks good. :)

MusikAnimal:
¡gracias!

PR #340, opened 2019-10-28
AutoEdits: Expand tools definitions

DannyS712:
See, eg, https://en.wikipedia.org/w/index.php?title=Talk:Embassy_of_China,_Pyongyang&diff=prev&oldid=912556387
Bug: T222323

### DIFF
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -354,7 +354,7 @@ parameters:
         Page curation:
             regex: 'using \[\[Wikipedia:Page Curation\|Page Curation'
             link: w:en:Wikipedia:Page_Curation/Help
-            namespaces: [0, 2, 3]
+            namespaces: [0, 1, 2, 3]
         Page swap:
             regex: '\/pageswap\|pageswap\]\]'
             link: WP:PAGESWAP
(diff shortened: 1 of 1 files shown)


config/semi_automated.yml
```diff
@@ -449,6 +453,11 @@ parameters:
             tag: 'OAuth CID: 860'
             link: User:Enterprisey/YABBR
             namespaces: [0]
+    en.wikisource.org:
+        Change status:
+            link: User:DannyS712/Change_status
+            regex: \[\[User:DannyS712/Change status\|Change status\.js\]\]
+            namespaces: [104]
```
MusikAnimal, line 460: I'm OK to merge this because en.wikisource's config is already so small, but in general we shouldn't be adding scripts to AutoEdits that have very low usage. Each and every script added slows down the tool a tad bit more.

MusikAnimal (approved):
Much appreciated!

config/semi_automated.yml
DannyS712: Understandable - I figured because of the few scripts already given, as well as the namespace restriction, it wouldn't be too much of an issue

PR #342, opened 2019-11-05
AutoEdits: Add 2 tools, tweak some others

DannyS712:
Commit 1: Update page curation with namespace 4 and pagetriage tag
Commit 2: Add userRightsManager for enwiki, in namespaces 3 and 4
Commit 3: Add Replacer tool for eswiki
Commit 4: Expand wikidata merge tool to include Special:MergeItems
Commit 5: Combine different commons global replace tools into 1 global rule
Commit 6: Rename Sagittarius to Capricorn, update regex and link

Bug: T237217
Bug: T237174
Bug: T237076
Bug: T231709
Bug: T236719
Bug: T237458

### DIFF
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -463,6 +463,10 @@ parameters:
             link: User:DannyS712/Change_status
             regex: \[\[User:DannyS712/Change status\|Change status\.js\]\]
             namespaces: [104]
+    es.wikipedia.org:
+        Replacer:
+            tag: 'OAuth CID: 498'
+            link: Usuario:Benjavalero/Replacer
     fa.wikipedia.org:
         Typo fixer:
             label: 'ویکی‌پدیا:اشتباه‌یاب'
--- a/config/semi_automated.yml
+++ b/config/semi_automated.yml
@@ -436,6 +436,10 @@ parameters:
             regex: '\(\[\[User:Enterprisey\/undo-last-edit\.js\|undo-last-edit\]\]\)'
             link: User:Enterprisey/undo-last-edit
             revert: true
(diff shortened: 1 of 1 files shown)


config/semi_automated.yml
```diff
@@ -520,9 +523,9 @@ parameters:
             regex: 'MediaWiki_talk:Gadget-labelLister.js\|labelLister'
             link: MediaWiki talk:Gadget-labelLister.js
             namespaces: [ 0, 120 ]
-        Merge.js:
-            regex: '\[\[MediaWiki:Gadget-Merge\.js\|merge\.js\]\]'
-            link: MediaWiki talk:Gadget-Merge.js
+        Merge:
+            regex: '\[\[MediaWiki:Gadget-Merge\.js\|merge\.js\]\]|^Merged Item'
```
MusikAnimal: I had to query the database to find out, but looks like the summaries for automatic merges are actually like:
```
/* wbmergeitems-to:0||Q5610520 */
```
or
```
/* wbmergeitems-from:0||Q25416796 */
```
This I'm sure is for localization purposes.

Something like `\/* wbmergeitems` should work.

MusikAnimal (changes requested):
Just one issue with Merge, everything else looks great!

The Sagittarius/Capricorn change got me thinking... I wonder if we should include an `aliases` option, so that older incoming links still work. Not important right now.

config/semi_automated.yml
DannyS712: Done (except I used `\/\* wbmergeitems`)

MusikAnimal (approved):
Thank you as always!

PR #351, opened 2020-01-13
EditCounter: show all 24h in timecard, and avg edits

MusikAnimal:
Bug: T241539

### DIFF
--- a/templates/editCounter/timecard.html.twig
+++ b/templates/editCounter/timecard.html.twig
@@ -45,7 +45,7 @@
     </table>
 </div>
 
-<canvas id="timecard-bubble-chart"></canvas>
+<canvas id="timecard-bubble-chart" height="100"></canvas>
 
 <script type="text/javascript">
     // MySQL uses ODBC standard, treating the first day (1) as Sunday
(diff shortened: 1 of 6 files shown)


templates/editCounter/timecard.html.twig
```diff
@@ -122,21 +132,28 @@
...
+                displayColors: false,
                 callbacks: {
+                    title: function(items, data) {
+                        return days[7 - items[0].yLabel + 1] + ' ' + items[0].xLabel + ':00';
+                    },
                     label: function(item) {
-                        // yLabels are reversed, and zero-indexed
-                        return days[7 - item.yLabel + 1] + ' ' + item.xLabel + ':00';
+                        var numEdits = [timeCardDatasets[item.datasetIndex].data[item.index].value];
+                        return`${numEdits} ${$.i18n('num-edits', [numEdits])}`;
```
ricordisamoa, line 156: May I ask why is there hard to translate string interpolation instead of reusing the parameter inside a message like `$1 {{PLURAL:$1|edit|edits}}`? 🤔

MusikAnimal: In the Twig templates, we format the first number (e.g. add commas), which wouldn't allow us to use it in PLURAL. JS doesn't have the same formatting. Anyway, I know we could just have $2 for the formatted number, and $1 for use in PLURAL, but it's been this way since the beginning, so I just followed the same pattern https://github.com/x-tools/xtools/blob/e2569fa563758b795fc9e7ba0ae472de410086f1/i18n/en.json#L264-L287

PR #353, opened 2020-02-29
GlobalContribs: Only query when actor id exists

DannyS712:
If it wasn't retrieved, then the user has no live edits
Bug: T246522


MusikAnimal (approved):
Thank you!! This works perfectly.

PR #356, opened 2020-04-02
AutoEdits: Add StubSorter on enwiki

supertassu:
Bug: T248077


MusikAnimal (approved):
Gosh, sorry I didn't notice this until now! Works perfectly :) I think alphabetically (case insensitive) this should go below "stubsearch", but I can change that in a different commit. Also FYI, the build system has been broken for a while. The tests actually do pass.

Merging. Thank you for your contribution!

PR #357, opened 2020-04-24
PageRepository::getWikidataInfo - update query

DannyS712:
Update query following removal of the wb_terms table
Bug: T250951

### DIFF
--- a/src/AppBundle/Repository/PageRepository.php
+++ b/src/AppBundle/Repository/PageRepository.php
@@ -281,10 +281,10 @@ class PageRepository extends Repository
 
         $sql = "SELECT wby_name AS term, wbx_text AS term_text
                 FROM $wdp.wbt_item_terms
-                LEFT JOIN $wdp.wbt_term_in_lang ON wbit_term_in_lang_id = wbtl_id
-                LEFT JOIN $wdp.wbt_type ON wbtl_type_id = wby_id
-                LEFT JOIN $wdp.wbt_text_in_lang ON wbtl_text_in_lang_id = wbxl_id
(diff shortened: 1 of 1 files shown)


src/AppBundle/Repository/PageRepository.php
```diff
@@ -275,14 +275,19 @@ public function getWikidataInfo(Page $page): array
...
-                AND term_type IN ('label', 'description')
-                AND term_language = :lang";
+        $wdp = 'wikidatawiki_p';
+
+        $sql = "SELECT wby_name AS term, wbx_text AS term_text
+                FROM $wdp.wbt_item_terms
+                LEFT JOIN $wdp.wbt_term_in_lang ON wbit_term_in_lang_id = wbtl_id
+                LEFT JOIN $wdp.wbt_type ON wbtl_type_id = wby_id
+                LEFT JOIN $wdp.wbt_text_in_lang ON wbtl_text_in_lang_id = wbxl_id
+                LEFT JOIN $wdp.wbt_text ON wbxl_text_id = wbx_id
```
matejsuchanek: Previous query had no LEFT JOIN's, so dependent callers may or may not expect NULL's.

matejsuchanek (changes requested):

src/AppBundle/Repository/PageRepository.php
DannyS712: Switching to just join

MusikAnimal (approved):
Works beautifully and is very fast. Thank you very much! It would have taken me a while to figure this out.

Don't mind the build failure, by the way. For the past months or so it always times out on the first run.

PR #364, opened 2020-09-07
Add more CropTool tags per T262124

mdaniels5757:
^


MusikAnimal:
Thank you!

FYI -- you can ignore test failures for right now. CI has been broken for a while... but I'm going to get it fixed soon. I tested this locally and it looks good to me. Merging.

PR #368, opened 2020-09-12
AutoEdits: add TranslationAdder for en.wikisource

MarcoAurelioWM:
Adding the TranslationAdder tool for en.wikisource.org

Bug: T250615


MusikAnimal:
Works in my testing. Thank you!

FYI, starting some months ago the initial CI build always fails for mysterious reasons. I still need to fix that. Maintainers are also the only ones who can re-trigger it, unfortunately. So for now, just ignore it :)

PR #369, opened 2020-09-22
AutoEdit: Fix regex for en.wiktionary script

MarcoAurelioWM:
Follow-up: 17eb31c; expands the regex to make it more specific and avoid false positives

Bug: T250615


MarcoAurelioWM:
@MusikAnimal This fixes a previous commit. Could you please take a look when you can? Thanks.

MusikAnimal:
Thank you!

PR #372, opened 2020-11-05
AutoEdits: move the semi_automated.yml config to on-wiki JSON page

MusikAnimal:
Bug: T262631


src/AppBundle/Helper/AutomatedEditsHelper.php
```diff
@@ -65,13 +78,42 @@ public function isAutomated(string $summary, Project $project): bool
...
+            return $this->cache->getItem($cacheKey)->get();
+        }
+
+        $title = 'MediaWiki:XTools-AutoEdits.json' . ($useSandbox ? '/sandbox' : '');
+        $ret = json_decode(file_get_contents(
+            $this->metaProject->getUrl() . "w/index.php?action=raw&ctype=application/json&title=$title"
+        ), true);
+
+        if (!$useSandbox) {
+            $cacheItem = $this->cache
```
DannyS712, line 92: Is there any way to invalidate this when the page is edited?

src/AppBundle/Repository/AutoEditsRepository.php
```diff
@@ -49,6 +65,19 @@ public function getTools(Project $project, $namespace = 'all'): array
...
+    /**
+     * Overrides Repository::setCache(), and will not call the parent (which sets the cache) if using the sandbox.
+     * @inheritDoc
+     */
+    public function setCache(string $cacheKey, $value, $duration = 'PT20M')
+    {
+        if ($this->useSandbox) {
+            return $value;
+        } else {
+            return parent::setCache($cacheKey, $value, $duration);
```
DannyS712: No need to wrap in an `else` block

templates/autoEdits/result.html.twig
```diff
@@ -216,7 +216,7 @@
                     </div>
                 {% endif %}
                 <p style="clear:both">
-                    <a target="_blank" href="https://phabricator.wikimedia.org/maniphest/task/create/?title=Request to add tool to AutoEdits&amp;projects=XTools&amp;description=Project: {{ project.domain }}%0A%0A{{ msg('auto-edits-request-name') }}:%0A%0A{{ msg('auto-edits-request-examples') }}:%0A%0A{{ msg('auto-edits-request-doc') }}:">
+                    <a target="_blank" href="https://meta.wikimedia.org/w/index.php?title=MediaWiki_talk:XTools-AutoEdits.json&action=edit&section=new&preload=MediaWiki_talk:XTools-AutoEdits.json/Preload">
```
DannyS712, line 219: Is there also a link to some documentation about how it is now on meta?

src/AppBundle/Helper/AutomatedEditsHelper.php
MusikAnimal: Not without querying the db or API (probably would need to be the letter) on every request. Not worth it in my opinion, as we're only talking up to a 20 min wait.

templates/autoEdits/result.html.twig
MusikAnimal: The thought is requests will mostly come from non-engineers, so I didn't want to confuse them with all the documentation. All maintainers need to know is those 4 questions asked in the preload template. Let's talk more on phab so others can chime in; I've got this along with a number of other things I wanted to ask you all about.

src/AppBundle/Helper/AutomatedEditsHelper.php
```diff
@@ -24,13 +27,23 @@ class AutomatedEditsHelper
     /** @var ContainerInterface The service container. */
     private $container;
 
+    /** @var Project $metaProject */
+    private $metaProject;
+
+    /** @var CacheItemPoolInterface The cache. */
```
DannyS712: "The cache." doesn't really help document anything since the variable is named $cache and the type is clearly a cache

src/AppBundle/Helper/AutomatedEditsHelper.php
```diff
@@ -24,13 +27,23 @@ class AutomatedEditsHelper
     /** @var ContainerInterface The service container. */
     private $container;
 
+    /** @var Project $metaProject */
```
DannyS712: $metaProject shouldn't be repeated in the documentation

src/AppBundle/Helper/AutomatedEditsHelper.php
```diff
@@ -24,13 +27,23 @@ class AutomatedEditsHelper
...
      * AutomatedEditsHelper constructor.
      * @param ContainerInterface $container
+     * @param Project|null $metaProject Solely used to mock the meta.wikimedia Project in AutomatedEditsTest
      */
-    public function __construct(ContainerInterface $container)
+    public function __construct(ContainerInterface $container, ?Project $metaProject = null)
     {
         $this->container = $container;
+        $this->cache = $container->get('cache.app');
+        $centralAuthProject = $this->container->getParameter('central_auth_project');
```
DannyS712: Suggest referencing $container directly, like on line 44, instead of $this->container; same applies to the use of $this->container on line 46

src/AppBundle/Helper/AutomatedEditsHelper.php
DannyS712: (noting here what I said on IRC) I meant a way to purge the cache from within the xtools interface

src/AppBundle/Helper/AutomatedEditsHelper.php
```diff
@@ -65,13 +72,45 @@ public function isAutomated(string $summary, Project $project): bool
         return (bool)$this->getTool($summary, $project);
     }
 
+    /**
+     * Fetch the config from https://meta.wikimedia.org/wiki/MediaWiki:XTools-AutoEdits.json
+     * @param bool $useSandbox
```
DannyS712: Please document that $useSandbox results in a different config source - the current documentation suggest that the main XTools-AutoEdits.json is always used

src/AppBundle/Repository/AutoEditsRepository.php
```diff
@@ -20,6 +20,22 @@ class AutoEditsRepository extends UserRepository
...
+    private $tags;
+
+    /**
+     * AutoEditsRepository constructor. Used solely to set $useSandbox (from AutomatedEditsController).
+     * @param bool $useSandbox
+     */
+    public function __construct(bool $useSandbox = false)
+    {
+        $this->useSandbox = $useSandbox;
+        parent::__construct();
```
DannyS712, line 35: Not sure about within this repo, but generally in my experience the parent::__construct() call is put first

src/AppBundle/Repository/AutoEditsRepository.php
```diff
@@ -61,7 +90,7 @@ public function getTools(Project $project, $namespace = 'all'): array
     public function countAutomatedEdits(Project $project, User $user, $namespace = 'all', $start = '', $end = ''): int
     {
         $cacheKey = $this->getCacheKey(func_get_args(), 'user_autoeditcount');
-        if ($this->cache->hasItem($cacheKey)) {
+        if ($this->cache->hasItem($cacheKey) && !$this->useSandbox) {
```
DannyS712: here and below, suggest putting the useSandbox check first to avoid checking the cache unnecessarily, and to be consistent with the order in AutomatedEditsHelper.php (or, if checking second is intentional, suggest being consistent with that)

src/AppBundle/Repository/AutoEditsRepository.php
```diff
@@ -278,19 +307,14 @@ public function getAutomatedEdits(
...
      *                  ]
      */
-    public function getToolCounts(
-        Project $project,
-        User $user,
-        $namespace = 'all',
-        $start = '',
-        $end = ''
-    ): array {
+    public function getToolCounts(Project $project, User $user, $namespace = 'all', $start = '', $end = ''): array
```
DannyS712, line 310: personally, I would prefer to have such long method declarations be split onto multiple lines, but either way suggest doing separately from this patch (as far as I can tell this is a whitespace only change, so it should have no impact on the result and just makes the diff a bit harder to review)

src/AppBundle/Repository/AutoEditsRepository.php
```diff
@@ -323,13 +347,12 @@ public function getToolCounts(
      * Get SQL for getting counts of known automated tools used by the given user.
      * @see self::getAutomatedCounts()
      * @param Project $project
-     * @param User $user
```
DannyS712, line 326: If the user parameter is unused, it should be removed, but how is this related to the rest of the patch? Suggest doing separately

tests/AppBundle/Helper/AutomatedEditsTest.php
```diff
@@ -32,7 +32,6 @@ public function setUp(): void
         $client = static::createClient();
         $container = $client->getContainer();
         $this->aeh = new AutomatedEditsHelper($container);
-        $this->cache = $container->get('cache.app');
```
DannyS712, line 35: Same as some of my other comments - how is this related to this patch / can it be done separately?

src/AppBundle/Helper/AutomatedEditsHelper.php
MusikAnimal: Hehe I agree. This is actually just to get around PHPCS from yelling at us. The sniffing rules are probably a little too strict. Anyway I expanded it to "The caching interface" :)

src/AppBundle/Repository/AutoEditsRepository.php
MusikAnimal: In production it makes virtually no difference, since it's in-memory APCu cache, but good practice nonetheless!

src/AppBundle/Repository/AutoEditsRepository.php
MusikAnimal: It's the XTools way, hehe :) By that I mean I largely work alone here, so little null fixes like this get combined with bigger PRs to save me time. Indeed $user is unused, I'll modify the commit message accordingly, same with below.

src/AppBundle/Helper/AutomatedEditsHelper.php
MusikAnimal: https://phabricator.wikimedia.org/T209334

MusikAnimal:
Tests. Finally. Pass. I spent more time fixing that than I did on the actual code changes, 10 times over...

Will get this on xtools-dev before merging

MusikAnimal:
Actually I am going to merge... just to sort of commit to it. Follow-up PRs to come as necesary.

MusikAnimal:
Also thank you again for the review :)

PR #379, opened 2021-02-24
Add optional dark mode

vitaly-zdanevich:


MusikAnimal:
Just letting you know I haven't forgotten about this! I have to finish up #378 first which is time-sensitive, then I'll be reviewing this :) Thank you for contributing!

assets/css/application.scss
```diff
@@ -923,3 +923,109 @@ input.form-control.show-loader {
...
+    table,
+    .panel-primary,
+    .panel-default,
+    th,
+    td,
+    .stat-list--group > tr:first-child > td,
+    hr,
+    .form-label,
+    .form-control {
+        border: none !important;
```
MusikAnimal: This makes adjacent input fields seem like they are one big textarea, and to me it's a little hard to read some tables without the borders. What do you think? Maybe try a darker shade of gray than the default #ddd?

MusikAnimal (commented):
Overall this looks amazing! Thank you! I have just one tiny concern, which might boil down to preference.

assets/css/application.scss
vitaly-zdanevich: Fixed, please check.

vitaly-zdanevich:
Screenshot of current state:

image

MusikAnimal:
Looks fantastic to me! We can iterate on it more as needed. Merging.

I'll get this deployed soon. Thanks again :)

mik2:
A few clearly missing things:

- grids in charts,
- borders in sections,
- table grids,
- alternating table backgrounds.

Moreover IMO font colors should be lighter to be more distinguishable from background.

x-tools dark mode

vitaly-zdanevich:
@MusikAnimal what do you think?

MusikAnimal:
It all sounds good to me! @vitaly-zdanevich @mik2 If either of you want to make a follow-up PR I'll happily review it, otherwise I'll try to get to it soon (I admit I don't use dark mode myself).

The only issue is with the charts. Those are built with [chart.js](https://www.chartjs.org/), specifically [version 2.7.0](https://github.com/chartjs/Chart.js/releases/tag/v2.7.0), which may or may not have all the customization options we need. In any case, we need to [detect `prefers-color-scheme:dark`](https://stackoverflow.com/a/57795495/604142) from the JavaScript to conditionally change the colors. It's probably not so simple. The easier option to is to give `<canvas>` elements a lighter background, but that might mean it doesn't look so "dark" anymore.

MusikAnimal:
> The only issue is with the charts.

Or actually, we could try inverting the colors of `<canvas>` elements in the same way we do "cheap" dark mode:
```
filter: invert( 1 ) hue-rotate( 180deg );
```
That looks semi-decent, I think. You'll have to make sure to apply that CSS to `.color-icon` elements, too, since those are used in chart legends (i.e. the table to the left of the chart in the Edit Counter's "Namespace Totals" section).

vitaly-zdanevich:
> The only issue is with the charts. Those are built with [chart.js](https://www.chartjs.org/), specifically [version 2.7.0](https://github.com/chartjs/Chart.js/releases/tag/v2.7.0), which may or may not have all the customization options we need. In any case, we need to [detect `prefers-color-scheme:dark`](https://stackoverflow.com/a/57795495/604142) from the JavaScript to conditionally change the colors. It's probably not so simple. The easier option to is to give `<canvas>` elements a lighter background, but that might mean it doesn't look so "dark" anymore.

Created issue https://github.com/chartjs/Chart.js/issues/8658

I think that our JavaScript can easy detect for dark mode https://stackoverflow.com/questions/59621784/how-to-detect-prefers-color-scheme-change-in-javascript

vitaly-zdanevich:
Add pull request to return grids to charts https://github.com/x-tools/xtools/pull/381

vitaly-zdanevich:
About customization of chart.js: I found this documentation https://www.chartjs.org/docs/latest/charts/bubble.html, I am not sure if it is better than `filter()` in CSS that is one line only.

mik2:
> > Moreover IMO font colors should be lighter to be more distinguishable from background.
> 
> Do you mean all fonts of from some specific places?

Generally all. Comparison to the colors of github dark mode looks like this:

```
         x-tools github
primary   b9b9b9 c7cfd7
secondary 767676 8a939d
link      3379b5 57a5fd
```

Moreover, charts in dark mode use currently secondary font color, which is not true for light mode (where it is something between primary and secondary). They could use something around current primary color.

vitaly-zdanevich:
@mik2 what do you think now? Is it ok?

mik2:
@vitaly-zdanevich you mean the version that is now on production? It is better, but not ideal. Maybe I will try to fix it myself, when I  get more time.

vitaly-zdanevich:
Ок

PR #381, opened 2021-03-17
Charts: return back grids for dark mode

vitaly-zdanevich:
See the first pull request with discussions https://github.com/x-tools/xtools/pull/379

### DIFF
--- a/assets/css/application.scss
+++ b/assets/css/application.scss
@@ -1035,7 +1035,8 @@ input.form-control.show-loader {
         border: 0;
     }
 
-    canvas {
+    canvas,
+    .color-icon {
         background: #fff;
         filter: invert() hue-rotate(180deg);
     }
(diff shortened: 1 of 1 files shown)


vitaly-zdanevich:
Before:

image


After:

image

assets/css/application.scss
```diff
@@ -1035,4 +1035,10 @@ input.form-control.show-loader {
         border: 0;
     }
 
+    canvas,
+    .color-icon {
+        background: #fff;
+        filter: invert() hue-rotate(180deg);
```
MusikAnimal: If I'm understanding it correctly, the [spec](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/invert()) requires that `invert()` be given an amount (`1` in this case). It might be coincidence that your browser assumed a value of `1`.

MusikAnimal (changes requested):
Definitely an improvement! The font color has also brightened which looks better, I think.

Just one tiny concern about the syntax of `invert()`

PR #405, opened 2022-01-23
Fix articleinfo.js.twig

tgr:
Can't look up an ID before inserting it into the document. Broken all the way back in a21d71fbab8ade3575f263fb979fc279ad022c99.

(did not test but change is trivial)

### DIFF
--- a/templates/articleInfo/articleinfo.js.twig
+++ b/templates/articleInfo/articleinfo.js.twig
@@ -12,10 +12,11 @@ $(function () {
         return;
     }
 
-    var markup = "<div id='xtools' style='font-size:84%; line-height:1.2em;" +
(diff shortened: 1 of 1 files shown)


MusikAnimal:
Thank you for this! The gadget hasn't been copied/pasted to the wiki in some time, but if were to have been it would obviously have broken without this patch.
Re https://phabricator.wikimedia.org/T299855 -- since XTools is now a WMF-only product (it used to work on any MW installation but no one used it), we don't really need this whole Twig system to serve what JavaScript people should use for the on-wiki gadget. So in case you wanted to work on T299855, feel free to write a gadget in your userspace or whatever and I can test then copy it to https://www.mediawiki.org/wiki/XTools/ArticleInfo.js

PR #431, opened 2022-11-17
i18n: Use number format also in pageviews offset

ebraminio:
Number format provides digits suitable for specific locales also in addition to addition of thousand separators so let's use it here also.

For example, https://xtools.wmflabs.org/api/page/articleinfo/fa.wikipedia.org/تهران?format=html&uselang=fa there is `30 روز` and this change should turn it `۳۰ روز` which will be correct for Persian.


MusikAnimal:
Thanks for this, and sorry for taking so long to review!

FYI, I probably won't get this deployed for another week or two, since we need to build a new VM for XTools and do some extensive testing.

PR #441, opened 2023-03-12
Add dockerfile

JackPotte:
Per https://phabricator.wikimedia.org/T280734.

To start:
- docker compose -f docker-compose.yml -f docker-compose.override.yml up -d
- cp .env.dev .env
- docker compose exec php bash
  - composer install
  - bin/console doctrine:database:create
  - bin/console doctrine:migrations:migrate
  - ./node_modules/.bin/encore dev --watch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,48 @@
+FROM php:7.4-fpm
+
+RUN apt-get update && apt-get install -y \
+      libxml2-dev \
+      libicu-dev \
+      libcurl3-dev \
+      libsqlite3-dev \
+      libedit-dev \
+      zlib1g-dev \
+      libfreetype6-dev \
+      libjpeg62-turbo-dev \
+      libmemcached-dev \
+      libzip-dev \
+      libonig-dev \
+      libpq-dev
+RUN docker-php-ext-install \
+        intl \
+        pdo_mysql \
+        pdo_pgsql \
+        opcache \
+        bcmath \
+        soap \
+        sockets \
+        zip \
+     && docker-php-ext-configure gd \
+     && docker-php-ext-install -j$(nproc) gd
+
+RUN pecl install apcu \
+    && pecl install apcu_bc \
+    && docker-php-ext-enable apcu --ini-name 10-docker-php-ext-apcu.ini \
+    && docker-php-ext-enable apc --ini-name 20-docker-php-ext-apc.ini
+
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
+RUN curl -sS https://get.symfony.com/cli/installer | bash -s -- "--install-dir=/usr/local/bin"
+
+RUN apt-get update && apt-get install -y \
+    openssh-client \
+    git \
+    gnupg \
+    apt-transport-https
+
+RUN curl -sL https://deb.nodesource.com/setup_14.x | bash
+RUN apt-get install -y nodejs
+RUN mkdir /.npm && chown -R 1000:1000 "/.npm"
+
+WORKDIR /var/www/xtools
+
+CMD ["php-fpm"]
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,0 +1,30 @@
+version: "3.8"
+services:
+  traefik:
+    image: traefik:v2.5
+    command:
+      - --providers.docker
+      - --entryPoints.web.address=:80
+      - --entryPoints.websecure.address=:443
+    ports:
+      - "80:80"
+      - "443:443"
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+
+  adminer:
+    image: adminer
+    labels:
+      - traefik.enable=true
+      - traefik.http.routers.adminer.rule=Host(`adminer.localhost`)
+      - traefik.http.routers.adminer.entryPoints=web
+    depends_on:
+      - mariadb
+
+  php:
+    labels:
+      - traefik.enable=true
+      - traefik.http.routers.app.rule=Host(`xtools.localhost`)
+      - traefik.http.routers.app.entryPoints=web
+    volumes:
+      - .:/var/www/xtools
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,18 @@
+version: "3.8"
+services:
+  mariadb:
+    hostname: mariadb
+    image: mariadb
+    ports:
+      - "3306:3306"
+    environment:
+      MYSQL_ROOT_PASSWORD: xtools
+
+  php:
+    user: ${UID:-1000}:${GID:-1000}
+    hostname: php8.2-fpm
+    build: .
+    ports:
+      - "8000:80"
+    depends_on: 
+      - mariadb


.env.dev
```diff
@@ -0,0 +1,86 @@
+# This file is a "template" of which env vars need to be defined for your application
+# Copy this file to .env file for development, create environment variables when deploying to production
```
MusikAnimal: So, we already have a `.env.dist` and a `.env`, with the latter git-ignored. It seems the preferred way to do this is to have the "template" (or default values) be the `.env` file, and the git-ignored overrides go in `.env.local`. https://symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local I can fix this in a separate commit, so for this commit I guess you don't need this file at all? Or is there something specific to Docker in here that we need?

src/DoctrineMigrations/Version20170623203059.php
```diff
@@ -1,7 +1,7 @@
 <?php
 declare(strict_types = 1);
 
-namespace App\DoctrineMigrations;
+namespace Application\Migrations;
```
MusikAnimal: I'm not sure what these changes are for. Looks like maybe you added them by mistake? `Application\Migrations` doesn't match the PSR-0/PSR-4 project structure.

MusikAnimal (commented):
Amazing, thank you!! Just a few questions

.env.dev
MusikAnimal: I'm working on this now, first installing Symfony Flex. That is apparently needed along with symfony/dotenv to ensure the right .env file is used for each environment.

src/DoctrineMigrations/Version20170623203059.php
MusikAnimal: I see where this came from. doctrine.yaml has the wrong migration path. I will update that in my PR as well, which I'll hopefully merge today/tomorrow which will unblock this PR.

MusikAnimal:
I have now merged #443 which reworks the .env file usage, and moves the Doctrine migration classes to their recommended location. So I think you just need to remove the .env.dev file added with this PR, and the changes to the migration files, and hopefully we're all set. You may want to rebase against `main` as well.

Thanks again for helping out! We've been wanting a Docker container for a long time now.

.env.dev
JackPotte: Yes, this wasn't for Docker itself but to be able so start Doctrine migrations from the container without editing any file.

JackPotte:
The rebase is done, thanks to you.

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+        opcache \
+        bcmath \
+        soap \
+        sockets \
+        zip \
+     && docker-php-ext-configure gd \
+     && docker-php-ext-install -j$(nproc) gd
+
+RUN pecl install redis \
+    && docker-php-ext-enable redis
```
MusikAnimal: The docs still mention Redis, which I need to update (note it says "outdated"), but Redis isn't an available option for caching anymore. We only use APCu.

MusikAnimal (commented):
I ran into some issues following the instructions. `docker compose -f docker-compose.yml -f docker-compose.override.yml up -d` ran without any errors, but entering the container I see this:
```
I have no name!@php8:/var/www/xtools$
```
It should have a shell name, right?

Also, the `/var/www/xtools` directory appears to be empty. Does the repo need to be cloned first? Is it possible to automate that in our Dockerfile?

Finally, it's unclear what port the server is running on.

Apologies if I ask any stupid questions! If it wasn't obvious by now, I'm not incredibly savvy with Docker, hehe :)

JackPotte:
"I have no name!" is "Normal" because the I had set the UID 1000 to avoid the container to claim to the host machine root account every file it touches (this UID can be overrided by a Docker .env file).

JackPotte:
/var/www/xtools is empty because of my mistake in docker-compose.yml:
`$HOME/www/xtools:/var/www/xtools`
In my host machine, xtools was one of the repos I needed to access to from the same container, so I had put an absolute path instead of a relative one, and then it could only work on my PC. I'm fixing that immediately...

JackPotte:
I think this is now better for CLI.

Concerning the ports, as I said In Phabricator, we would need to add a vhost (ex: Apache or Nginx) to know which container to add to listen to 80 and 443 (and make Xtools work in our navigators).
MySQL is still on 3306.
I had put PHP on 8000 to allow the host machine or other containers (ex: the web server) to contact it through it in the future (but we can also force a fixed IP instead). NB: if these containers are in the same docker-compose, they can call it by its name too.

MusikAnimal:
So, I changed my `.env.local` to have the credentials you used in your `.env.dev` file (namely, the DATABASE_HOST is mariadb as opposed to localhost). That worked great and I was able to run the migrations.

I did get an error when trying to run `./node_modules/.bin/encore dev` however: `Node Sass does not yet support your current environment: Linux 64-bit with Unsupported runtime (93)`. I then tried running `npm install` and got:
```
npm ERR! code EACCES
npm ERR! syscall mkdir
npm ERR! path /.npm
npm ERR! errno -13
npm ERR! 
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR! 
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 1000:1000 "/.npm"
```

I tried running `chown` as suggested and two problems: there's no `sudo` on the VM (I think that's expected?), and the `/.npm` directory doesn't exist. Any suggestions on how to fix that?

I'm still confused about the server. I see _something_ is running on port 8000. Browsing to that I get a browser error "The connection was reset". I know you said you need to add Apache or Nginx (I think we'd prefer the former since that's what we use in production). Is that still needed for the server to work? When I work off of my local without Docker, I use the built-in Symfony server, which I believe is just a wrapper around the PHP default server. So I think it's OK if it's all single-threaded or whatever, if that means anything.

Thanks for your patience and working on this for us!

JackPotte:
Hello and sorry for the late response.

I've just fixed the `npm install` problem.

Concerning the HTTP server, one purpose of Docker is to provide the exact same configuration for one app in all of its environments. So, I think that using the Symfony server only locally would just add some complexity and potential undetected bugs in the environments which would not use it. For example, a typical one I've seen in my carrer was a difference in these variables, causing some crashes in prod:
```
client_max_body_size 32m;
fastcgi_read_timeout 600;
proxy_read_timeout 600;
```

So I've begun to look in SSH on the wmflabs.org server, in /data/project/xtools, and I didn't find any Apache or Nginx vhost but the config for https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web/Lighttpd instead.
Do you think this is the target for our configurations, especially since the Kubernetes server has become the norm: https://wikitech.wikimedia.org/wiki/Help:Toolforge/Kubernetes?
FYI, Lighttpd doesn't provide an official Docker image, so we should add in the docker-compose.yml the most used among https://hub.docker.com/search?q=lighthttpd

MusikAnimal:
> Hello and sorry for the late response.

No worries :) I've been very busy as well!

> Concerning the HTTP server, one purpose of Docker is to provide the exact same configuration for one app in all of its environments. So, I think that using the Symfony server only locally would just add some complexity and potential undetected bugs in the environments which would not use it.

If we get it running with Apache that'd be great, just so long as the profiler is still there: https://symfony.com/doc/current/profiler.html That is a very important development tool.

> So I've begun to look in SSH on the wmflabs.org server, in /data/project/xtools, and I didn't find any Apache or Nginx vhost but the config for https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web/Lighttpd instead. Do you think this is the target for our configurations, especially since the Kubernetes server has become the norm: https://wikitech.wikimedia.org/wiki/Help:Toolforge/Kubernetes? FYI, Lighttpd doesn't provide an official Docker image, so we should add in the docker-compose.yml the most used among https://hub.docker.com/search?q=lighthttpd

We're not on Toolforge anymore but on [Cloud VPS](https://symfony.com/doc/current/profiler.html). The instructions for setting up the server can be found at https://wikitech.wikimedia.org/wiki/Tool:XTools. Sorry I didn't mention that doc sooner! Most everything in that doc can probably be followed for the Docker container, but you can ignore the API server part, as well awstats, and the Apache config doesn't need to be quite as complicated. If it's too much work, I think it's totally fine to use the Symfony web server. You're right it's better to be close to prod as possible, but in our case there isn't much variance between dev and prod when it comes to writing code.

I wanted to also say that since you started this PR, I've reworked a lot of the code. That's all done now, so what's in `main` is stable.

ttloves2 (approved):
would love to work on this as well.

Justman100:
@MusikAnimal

What here now?

MusikAnimal:
All contributions are welcome! Feel free to fork the repo and borrow from this PR, since you can't push to it. I would prefer to ping @JackPotte once more to see if they're interested in finishing it up first, though.

The truth for me is that I'm a complete dummy when it comes to Docker :/ I'm happy to help test, but won't be able to offer much help beyond that unfortunately.

JackPotte:
Hello, it's in my TODO list for one year (sorry about that)... So maybe next month.

PR #452, opened 2023-04-22
Update redirected link

Meno25:
* User was renamed

https://en.wikipedia.org/w/index.php?title=Special:Log&logid=127718833


MusikAnimal:
Thanks!

PR #483, opened 2024-07-03
Fix contrast of gray text in pageInfo API HTML response

jwbth:
image

#a2a9b1 [doesn't](https://www.siegemedia.com/contrast-ratio#%23fff-on-%23a2a9b1) have an appropriate contrast ratio with the white background. Use a color intended for secondary text, and while we're at it, use a CSS variable available in dark mode (I assume this HTML is supposed to be used in MediaWiki context, but if it is not, the color will fall back to #54595d).


MusikAnimal (approved):
Thank you! I'll get this deployed today

PR #484, opened 2024-08-20
EditCounter: local timezone option fix, accommodating fractional timezones

IndigoManedWolf:
Bubbles now move between rows (days) and change color when changing timezone, instead of looping in the same day. Fractional timezones (e.g. India Standard Time, UTC+5:30) no longer break graph (particularly when turning local timezone option on and off repeatedly).
Bug: T362818


assets/js/editcounter.js
```diff
@@ -343,7 +347,7 @@ xtools.editcounter.setupTimecard = function (timeCardDatasets, days) {
                     ticks: {
                         beginAtZero: true,
                         min: 0,
-                        max: 23,
+                        max: 24,
```
MusikAnimal, line 350: I think this might be why `24:00` is now showing up on the chart.

MusikAnimal (commented):
Amazing work!

Maybe I'm missing something, but is showing a column for `24:00` an intentional change? That's the only issue I'm seeing.

IndigoManedWolf:
> Amazing work!
> 
> Maybe I'm missing something, but is showing a column for `24:00` an intentional change? That's the only issue I'm seeing.

Intentional from when I was looking at what happens with fractional timezones, but I had forgotten by the time I made the PR. The motivation is this behavior:

If you set your timezone to a fractional timezone (such as UTC+5:30, where around 1.4 billion people live), bubbles move off of the edge of the chart
image

Actually, if there is a better approach you have in mind I can implement it.

MusikAnimal (commented):
Sorry for taking so long to get back to this!

> if there is a better approach you have in mind I can implement it

I thought there was a way to add padding to the chart, but I wasn't able to figure it out. We're using a rather old version of Chart.js that I don't think allows for this.

We can however at least hide the "24:00" label. That's really what was bothering me, as that is not a valid time. On line 356 of editcounter.js we could have:
```
if (value % 2 === 0 && value < 24) {
```
or something to that effect.

MusikAnimal:
The aforementioned isn't a big deal. I'm just going to merge :)

Thank you for taking the time to both identify this bug and fix it!

PR #495, opened 2025-04-14
EditCounter: fix two problems and a future problem with reblocks

agoujot:
1. (re)blocks actually doesn't count reblocks. One of the commits changes blocksOnly to include reblocks (and so exclude only unblocks), as it's only used for this purpose.
2. Currently, for infinite reblocks of non-infinite blocks, the code is made so that the longest block duration will be recorded as (time between start of block and reblock)-1 seconds, instead of infinite. One of these commits fixes that part of getLongestBlockSeconds so that it does return infinite for these cases
3. ipblocks is deprecated since MW 1.42 (June 2024); eventually it'll cause issues. The last commit switches the query to the new recommended tables. This may or may not be related to an issue with "current block" listed as "-" for many users who are blocked. If it is, then good; if it isn't, using deprecated stuff is never a good idea.

Note:
 - Although I've read a fair bit of PHP, I very rarely write some, so please check I haven't made syntax errors or other stuff like that;
 - And I'm not even remotely close to having a working testing environment, so I'd appreciate if someone else can test what I made.

Bug: T391824


MusikAnimal:
Thanks for this! I will review this soon. Looks like I need to fix CI, too.

agoujot:
For the table switch, specifically, I just corrected some mistakes I'd made (in the actual names), but I'm not sure what should be put in the last parameter of [getTableName](https://github.com/agoujot/xtools/blob/main/src/Repository/Repository.php#L212).

MusikAnimal:
Can you rebase (or merge in `main`) and push again, for all of your PRs? I just fixed CI finally :)

I hope to everything reviewed before this Friday.

Thanks for contributing! :smile:

MusikAnimal (approved):
:100: Thank you!!! I'm surprised these bugs have gone unnoticed for this long!

Don't worry about the tests, since you aren't able to set up a dev environment. I can add coverage in a follow-up.

MusikAnimal:
Also FYI, [multiblocks](https://www.mediawiki.org/wiki/Help:Manage_blocks) is coming soon (hence the schema change), so we'll need to rethink how we display the data. Only two wikis have it enabled right now, so it's not urgent.

agoujot:
Also, I was wondering: with what frequency is xtools.wmcloud.org updated? To see the effect notably if the table change had something to do with the issue with current blocks issue.

PR #496, opened 2025-04-16
Blame: and Authorship: Update wikiwho links

agoujot:
The f-squared link, [without](https://f-squared.org/wikiwho/) or [with](https://www.f-squared.org/wikiwho/) the www prefix, is now dead (and it doesn't exist anywhere on f-squared anymore).

So, I replaced those links by links to [the mediawikiwiki page](https://www.mediawiki.org/wiki/WikiWho), which looks like a good user-facing link to me.


agoujot:
Give this only replaced links in the code, I don't know what to make of the above bot message.

PR #497, opened 2025-04-16
EditCounter: make the day count in the timecard update with timezone change

agoujot:
Problem: when the user changes the timezone, currently the sums at the right don't get updated.

Solution: replacing timeCardDatasets in the callback line 339 with chart.data.datasets.

timeCardDatasets is the initial value, whereas chart.data.datasets is the updated value (timeCardDatasets doesn't get updated because we do chart.data.datasets = chart.data.datasets.map(...))

(And this time it's in a language I've written a lot in.
And of course, I forgot to link to the bug in the commit again...)

Bug: T385962


MusikAnimal:
There's a build step that needs to be ran locally. Let me see if I can take care of that for you.

assets/js/editcounter.js
```diff
@@ -336,7 +336,7 @@ xtools.editcounter.setupTimecard = function (timeCardDatasets, days) {
                             if (index === 0 || index > 7) {
                                 return '';
                             }
-                            return timeCardDatasets[index - 1].data.reduce(function (a, b) {
+                            return chart.data.datasets[index - 1].data.reduce(function (a, b) {
```
MusikAnimal: This errors out on the first run:
```
ReferenceError: chart is not defined
```
Perhaps check against `window.chart` first.

I hate there's so much usage of the global scope, but that's a preexisting problem I created years ago :-P

MusikAnimal (dismissed):
Hopefully an easy fix.

agoujot:
This is weird. I was afraid of such issues, so I made a quick test in the console:
```
window.test = [ { fun : () => console.log(test[1].data) }, { data:"hello!" } ];
test[0].fun()
```
And it worked, it did output `hello`.

Perhaps the callback is called during initalisation of the variable, too.

Eh, whatever. `window.chart ? chart.data.datasets : timeCardDatasets` should work fine.

assets/js/editcounter.js
agoujot: Should be fixed by the use of `window.chart ? chart.data.datasets : timeCardDatasets` instead.

If that *still* doesn't work, we can always redefine the entire callback when we update, but it'd be very messy.

MusikAnimal (changes requested):
No longer errors, but unfortunately it doesn't work. The values on the right stay the same. I did a little debugging and it seems `chart.data.datasets` apparently isn't updated (?)

I can try to debug more, but this might be easier if we can get your dev environment set up. I'm happy to help with that. If you a Wikimedia developer account, and you've used Toolforge before (i.e. have SSH keys set up), you're already most of the way there. Development docs are at https://www.mediawiki.org/wiki/XTools/Development

agoujot:
Yeah, I really should take time and set up a dev environment.

Reading the list for the first time, I think I have everything except the git setup. But I'll probably manage.

agoujot:
I'd appreciate if you could lend a hand. I *think* I've done everything right, but I'm getting an exception: `SQLSTATE[HY000] [2002] Connection refused`.

With `mariadb` I  get `ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)`. What exactly should I do? I found a lot of conflicting answers out there.

Being on a weird linux distribution and not having updated my package list for a year or two doesn't help, but normally I have versions >= the requirements.

MusikAnimal:
> I'd appreciate if you could lend a hand. I _think_ I've done everything right, but I'm getting an exception: `SQLSTATE[HY000] [2002] Connection refused`.
> 
> With `mariadb` I get `ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)`. What exactly should I do? I found a lot of conflicting answers out there.

For Linux, I think you should be able to get by with just the PHP adapter: `apt install php-mysql` (or `php8.1-mysql`, whatever version matches your local PHP version). You don't actually need a MySQL server running, since you'll want to [develop against the replicas](https://www.mediawiki.org/wiki/XTools/Development#Developing_against_Toolforge_replicas).

We can also chat in realtime, either on [IRC](ircs://irc.libera.chat:6697/wikimedia-xtools) or Discord. I'm MusikAnimal in both places.

We have a two-year old PR to create a Docker container (#441). I wish I knew enough about Docker to get that finished…

agoujot:
(Left a friend request at discord.)

On `SQLSTATE[HY000] [2002] Connection refused`:  I suppose it's an issue with my replica config? It's

```
DATABASE_HOST=127.0.0.1
DATABASE_PORT=
DATABASE_NAME=
DATABASE_USER=
DATABASE_PASSWORD=

DATABASE_REPLICA_USER=[usernameexactlyasinreplica.my.cnf]
DATABASE_REPLICA_PASSWORD=[passwordexactlyasinreplica.my.cnf]
DATABASE_REPLICA_HOST_S1=127.0.0.1
DATABASE_REPLICA_PORT_S1=4711
DATABASE_REPLICA_HOST_S2=127.0.0.1
DATABASE_REPLICA_PORT_S2=4712
DATABASE_REPLICA_HOST_S3=127.0.0.1
DATABASE_REPLICA_PORT_S3=4713
DATABASE_REPLICA_HOST_S4=127.0.0.1
DATABASE_REPLICA_PORT_S4=4714
DATABASE_REPLICA_HOST_S5=127.0.0.1
DATABASE_REPLICA_PORT_S5=4715
DATABASE_REPLICA_HOST_S6=127.0.0.1
DATABASE_REPLICA_PORT_S6=4716
DATABASE_REPLICA_HOST_S7=127.0.0.1
DATABASE_REPLICA_PORT_S7=4717
DATABASE_REPLICA_HOST_S8=127.0.0.1
DATABASE_REPLICA_PORT_S8=4718

DATABASE_TOOLSDB_HOST=127.0.0.1
DATABASE_TOOLSDB_PORT=4720
```
having done
```
ssh -f -L 4711:s1.web.db.svc.wikimedia.cloud:3306 -L 4712:s2.web.db.svc.wikimedia.cloud:3306 -L 4713:s3.web.db.svc.wikimedia.cloud:3306 -L 4714:s4.web.db.svc.wikimedia.cloud:3306 -L 4715:s5.web.db.svc.wikimedia.cloud:3306 -L 4716:s6.web.db.svc.wikimedia.cloud:3306 -L 4717:s7.web.db.svc.wikimedia.cloud:3306 -L 4718:s8.web.db.svc.wikimedia.cloud:3306 -L 4720:tools.db.svc.eqiad.wmflabs:3306 -N alien4444@login.toolforge.org
```
to setup the tunnels (I added -f to make it run in the background, else it actually ssh'd into toolforge and there I can't install      anything, and -N because else it complains).

agoujot:
There we go! Found the issue: even if we change the y property of an hour bubble, it's still classified within its original day; so I just filtered all hours to keep only those that in the end land in the right day. As we've got a fixed N of 7*24=168 that really isn't big, it should be fine.

agoujot:
On the compiled assets, though: I'm not sure what to upload and how. e.g. for the vaguely hash-like string at the end of the app.js filename, IDK where I'm supposed to get that from.

MusikAnimal (approved):
Thank you! :smile: 

n.b. use `npm run build` to compile the assets before committing, which will remove the old ones automatically. I know this is annoying but we don't have a build pipeline for deployments yet.

You can also use `npm run watch` during development, so you don't have to manually rebuild constantly.

agoujot:
I did figure it out eventually; my issue is I haven't figured out how to commit with git yet (yes, really :clown_face:), so I'm using the github UI that only lets you change a few files at a time.

PR #498, opened 2025-04-16
EditCounter: Handle partially deleted rights log

agoujot:
Knowing that:
 - logs can only be partially revdeleted.
 - log_params is only null when revdeleted (at any rate in the whole of enwiki_p)
 - log_params is always null when *any* revdeleting was used
 - revdeleted values are null in the return

From that:
 1. in userrights.php: we here leave nulls instead of skipping for partially revdeleted logs. This includes added and removed.
 2. in userrightsrepository.php: we use LEFT OUTER JOIN and not JOIN actors table, to not exclude revdeleted actors. Also remove the IFNULL around comment_text because null is a good signal to twig.
 3. in rights_changes_table.html.twig and rights_changes_table.wikitext.twig: if users or comments or rights are null, put a more graceful message.
 

Things to consider:
 - This whole business is quite experimental, especially for the twig parts. I don't have the technical means to test any of my code for XTools.
 - I made the choice to use null to tell twig it's been deleted, but maybe something else should be used.
 - I used the message 'error-deleted-data-link' for revdeld information ("removed from public view" in english), but if maybe it wasn't meant for that
 - If it was a *decision* to leave out all partially revdel'd logs, then that should be documented in the user interface.

Bug: T391753


MusikAnimal (changes requested):
Causes the `TypeError`:
```
in_array(): Argument #2 ($haystack) must be of type array, null given in src/Model/UserRights.php (line 263)
```
This PR is complicated enough that I think you'd need a dev environment set up to verify your work, and ideally add a test case for these changes. If a dev environment doesn't work out or just isn't possible for you, rest assured I will still get this merged eventually, we'll just have to take a slower route where I do the debugging for you.

This is to say I'm _very_ grateful for any all contributions to XTools, so please don't let me discourage any further contributions (with or without a working dev environment)!

agoujot:
There we go! Should be done. In the end, twig didn't have a clean way of distinguishing null and [], so I ended up querying log_deleted too and directly giving bools to twig based on log_deleted. Tested html and wikitext output with /ec/en.wikipedia.org/Kencf0618 and /ec/en.wikipedia.org/Chew.

src/Model/UserRights.php
```diff
@@ -223,77 +226,86 @@ private function processRightsChanges(array $logData): array
...
-                            foreach (array_reverse($rightsChanges, true) as $timestamp => $change) {
-                                if (in_array($right, $change['removed']) && !in_array($right, $change['added']) &&
-                                    'automatic' === $change['grantType']
-                                ) {
-                                    unset($rightsChanges[$timestamp]);
+                // As log_params is NULL, we don't know.
+                // Leave arrays here to not crash later.
+                // Twig will know from log_deleted.
+                $added = [];
+                $removed = [];
```
MusikAnimal, line 236: (nitpick) Consider using `continue;` here, then we don't need the `else {}` block and thus we don't have to indent those lines of code

templates/editCounter/rights_changes_table.html.twig
```diff
@@ -23,21 +23,33 @@
...
                 </td>
                 <td class="{{ change.grantType == 'pending' ? 'pending ' : '' }}sort-entry--rights" data-value="{{ change.added|length + change.removed|length }}">
-                    {% for right in change.added %}
-                        <div class="diff-pos">{{ ec.userRights.rightsName(right)|trim }}</div>
-                    {% endfor %}
-                    {% for right in change.removed %}
-                        <div class="diff-neg" dir="ltr">-{{ ec.userRights.rightsName(right)|trim }}</div>
-                    {% endfor %}
+                        {% if change.paramsDeleted %}
+                            <i class="text-muted">{{ msg('error-deleted-data-link') }}</i>
```
MusikAnimal: The message is used inside of another message, and hence some translations may not make sense in this isolated context. Fortunately we have the `edit-summary-removed` message which despite the name is used for this very purpose; It's documented to be identical to the [rev-deleted-comment](https://en.wikipedia.org/wiki/MediaWiki:Rev-deleted-comment) in MediaWiki, which also reads "edit summary removed".

MusikAnimal (changes requested):
Looks amazing! Thank you :)

In addition to the comments: Knowing you're not a PHP guru I won't pester you into writing new test coverage. For now we just need to fix the DB mocks in `UserRightsTest::testUserRightsChanges()` to return `log_deleted`. Simply adding `'log_deleted' => 0,` to each mock should do it, then also add `'paramsDeleted' => false` etc. to the expectations. I'm happy to help if you need it.

templates/editCounter/rights_changes_table.html.twig
agoujot: The last commits use error-deleted-data-link for all columns; perhaps then we want to change these too?  Looking into the details, we already have `username-removed`. Don't know really what to put for the rights changes proper, though.

src/Model/UserRights.php
agoujot: Yeah, it used (when I was trying to figure out what does what) to be more complex; agree `continue` makes more sense with what it ended up being.

agoujot: Hang on - no, actually, we can't use continue, else we're skipping the whole point of this :) (as it'll exclude the stuff at the very end of the loop that actually adds to the list).

templates/editCounter/rights_changes_table.html.twig
agoujot: There we go, done except for the params.

agoujot:
On test coverage: if you just tell me how it works :), glad to help. Or maybe I'll try to figure it out from the source.

src/Model/UserRights.php
MusikAnimal: Ah, right! Then moving to a `private function` is probably what I would do, then. But no need to bother, this passes the linter rules (which for this repo are quite strict) so beyond that is just a nitpick.

templates/editCounter/rights_changes_table.html.twig
MusikAnimal: Hmm, so 'tis a shame that the `logging` view for Toolforge will scrub out the `log_params` even when it's visible on-wiki! I think either the `unknown` or `data-unavailable` messages are better suited here, or we create a new message. `error-deleted-data-link` can't be used as it is a "lego" message (a message that's part of another message).

templates/editCounter/rights_changes_table.html.twig
```diff
@@ -23,21 +23,33 @@
...
                 <td class="{{ change.grantType == 'pending' ? 'pending ' : '' }}sort-entry--performer" data-value="{{ change.performer }}">
-                    {% if change.grantType == 'manual' %}
-                        {{ wiki.userLink(change.performer, change.type == 'local' ? project : metaProject) }}
+                    {% if change.performerDeleted %}
+                        <i class="text-muted">{{ msg('username-removed') }}</i>
+                    {% else %}
+                        {% if change.grantType == 'manual' %}
+                            {{ wiki.userLink(change.performer, change.type == 'local' ? project : metaProject) }}
+                        {% endif %}
                     {% endif %}
```
MusikAnimal, line 44: What happens for an automatic change with `performerDeleted`? I guess that can never happen?

I was going to suggest using `{% elseif` then noticed this potential code smell.

MusikAnimal:
> On test coverage: if you just tell me how it works :), glad to help. Or maybe I'll try to figure it out from the source.

What you have looks awesome, thank you!

MusikAnimal (changes requested):
Just need to change the `error-deleted-data-link` message and this is good to go, I think! My other comments are not blockers.

Thank you for the tests :)

templates/editCounter/rights_changes_table.html.twig
agoujot: I don't think an automatic change can be revdel'd at all, because admins plain don't get to see the change, so they don't have a revdel button. This is probably intentional. There can't be anything to revdel in an auto change (everyone knows when and how it happened and who did it and what was the comment).

agoujot: Hmm. Weird thing: there is a Special:Log option to see autopromotes, but it can't manage to find autoconfirmed, only other autopromotes. So it can theoretically be done. But then, if performer is deleted I think showing "user deleted" is what we should do, as it what the admins wanted. I checked, and `log_actor` is the user themselves for autopromotions. Note: thanks god testwiki exists. [We can see](https://testwiki.wiki/wiki/Special:Log?type=rights&user=&page=&wpdate=&tagfilter=&subtype=autopromote&wpFormIdentifier=logeventslist) that it is possible, and that what MW does is show "removed". So, we should do the same.

templates/editCounter/rights_changes_table.html.twig
agoujot: Done, used `data-unavailable`.

agoujot:
(note: I did use data-unavailable for both commits, just mixed up in second commit message.)

templates/editCounter/rights_changes_table.html.twig
MusikAnimal: Indeed, autoconfirmed is a special case! It goes by [$wgAutopromote](https://www.mediawiki.org/wiki/Manual:$wgAutopromote) which [does not actually add the user to a user group](https://www.mediawiki.org/wiki/Manual:$wgAutopromote#Caveats), as opposed to [$wgAutopromoteOnce](https://www.mediawiki.org/wiki/Manual:$wgAutopromoteOnce) which does. To work around this, we have to first parse the expression used for the `autoconfirmed` formula, resulting in this atrocity (including the use of [eval](https://www.php.net/manual/en/function.eval.php) :vomiting_face:):

https://github.com/x-tools/xtools/blob/7d6ff4805ded5995e45707b68bf556ef8f8feae7/src/Repository/UserRightsRepository.php#L178

Which reminds me, a colleague kindly added an API endpoint for this last year ([T196059](https://phabricator.wikimedia.org/T196059)) and I have yet to update XTools to make use of it. I've filed [T392551](https://phabricator.wikimedia.org/T392551)

Anyway, thanks for doing the investigative work here, and reminding me of this tech debt! :stuck_out_tongue:

MusikAnimal (approved):
:shipit:

PR #501, opened 2025-04-17
EditCounter: Use review not patrol stats for enwiki

agoujot:
(Ignore the first five commits, it's just a commit in the wrong branch, its revert, typos in the revert, and merging the main to here. I've cleaned my main branch so this shouldn't happen again.)

Enwiki is the only wiki to use pagetriage, and they barely use patrol.

Therefore, replace the patrol stats by the pagetriage-curation reviewed stats. They have requested this a certain number of times, e.g. T366397, https://www.mediawiki.org/wiki/Topic:Xxhmzabmfwnsuwck, &c.

Which means:
 - add reviewed to the list of log types where we want 0 if it's null;
 - add a public function that just gives the reviews
 - use {% if project.databaseName == enwiki %} in twig
 
NL said that "Page curation" works as a label and will probably not cause too much confusion, so replace the patrol message by that. (we've already got it translated.)

Bug: T366397


agoujot:
Ok, so, after testing and numbers checking:
 - corrected reviewed to pagetriage-curation-reviewed
 - added actions reviewed-article and reviewed-redirect (ah, nothing like outdated documentation. also, the above is not a superset of these. Possibly an old classification from before it said article/redirect? Anyhow the NPP folks count it separately).
 - added patrol-patrol count in that, as the enwp people count it in.(Asked confirmation, just in case.)

agoujot:
There we go. Should re ready.

templates/editCounter/general_stats.html.twig
```diff
@@ -340,6 +340,14 @@
                         {{ wiki.userLogLink(user, project, ec.approvals|num_format, 'review', 'accept') }}
                     </td>
                 </tr>
+                {% if project.databaseName == "enwiki" %}
```
MusikAnimal: This locks this to show only for enwiki, even if PageTriage gets installed elsewhere at a later time. Or it could be a 3rd party wiki with PageTriage installed (see also [T344090](https://phabricator.wikimedia.org/T344090)).

I think more ideally we'd use [Project::getInstalledExtensions()](https://github.com/x-tools/xtools/blob/0b47bf53a70751b79eeafb58941656cb93ac0b28/src/Model/Project.php#L255) to look for `PageTriage`, and pass i.e. `hasPageTriage` to the view.

MusikAnimal (commented):
Looking good! I'm happy to merge as-is, but would like to discourage hard-coding database names (though I know we are guilty of doing it in other parts of XTools).

>  added patrol-patrol count in that, as the enwp people count it in.(Asked confirmation, just in case.)

Looks like we're _not_ including `patrol-patrol` now? That's good because it would double count reviews, I think. Each initial `pagetriage-curation-reviewed*` I believe also produces a `patrol-patrol` log entry.

MusikAnimal:
Not required but for bonus points: update `EditCounterTest::testLogCounts()` accordingly :)

templates/editCounter/general_stats.html.twig
agoujot: Done!

agoujot:
On double-counting: not exactly. Some do, some don't, and anyhow after I asked they told me to split the two, so I did. So right now we've got the two next to each other and they do whatever they want to do with it.

Changed the database name hardcoding.

Also added tests for reviews.

MusikAnimal (approved):
Thank you! :100:

PR #505, opened 2025-04-22
Display time order correctly in RTL wikis

ebraminio:
This turns

image

to 

image

in RTL wikis but `<bdi>` wouldn't hurt the use in LTR wikis also.


MusikAnimal:
Thank you! :)

PR #504, opened 2025-04-22
EditCounter: Scale timeCard bubbles to chart height

agoujot:
To be precise, make the largest's radius (height-20)/9/2. -20 for the bottom labels, /9 because there are a bit more than eight sections the size of a bubble, and /2 to get radius not diameter.

Bug: T270984

PR #506, opened 2025-04-24
EditCounter: add hour totals to time card

agoujot:
Essentially, adding a first line before the hour labels that contains the totals (using the same timezone logic I used for the day counts).

As even in our seriously outdated chartjs, we can make multi-line labels with string arrays.

Bug: T251836


MusikAnimal (approved):
I'm willing to merge anyway, but the x-axis legend can now become illegible on smaller viewports:
Screenshot from 2025-04-24 19-30-46

Do you think that's fixable? It would be quite hacky, but we could listen for window resizes and show/hide the totals accordingly (I think the hour legends are more important).

Or maybe I should get a start on [T392531](https://phabricator.wikimedia.org/T392531) already, if bumping Chart.js would allow us to have an x-axis on both the top and bottom.

agoujot:
Will try to work something up. Normally we don't need to listen because chartjs listens; only got to hide some ticks in  afterTicks callback, like I did for the log scale.

A chartjs upgrade would solve pretty much all of this anyhow; they've done patches on diverse causes of overlapping. In this case it's [this one](https://github.com/chartjs/Chart.js/pull/10641).

Question: does the above screenshot include #504 ? If it does, I've got rechecking to do.

MusikAnimal:
> Question: does the above screenshot include #504 ? If it does, I've got rechecking to do.

I tested on b7187f9bd59be0a86ed7be624f0b129e433167b5 which looks like no, it does not include #504.

> A chartjs upgrade would solve pretty much all of this anyhow

Awesome! I'll get to work on https://phabricator.wikimedia.org/T392531

agoujot:
(I believe we shouldn't merge this until either I manage to patch something or we upgrade chartjs, because becoming illegible for everything below 900px is really meh. 900px is a lot.)

agoujot:
Well, after a lot of testing stuff and wondering about, I'm pretty convinced there's no clean way to hide some of these labels, that doesn't involve reimplementing at least a hundred lines of later versions of chartjs.

So, I just deactivated the hour totals on anything that's narrower than 1000px, as anyhow they couldn't have read it; and like this the hour labels are legible again on small screens.

Also, i'm taking the occasion to remove the "24:00" label, as anyhow it doesn't make much sense (there can't be any edits there), and it created a phantom 0 hour total.

agoujot:
Ah, just stumbled on the pull request where the 24:00 was added, to accomodate fractional timezones, else they get off the chart. 

Well, then, just going to readd that tick but hide the label.

PR #507, opened 2025-04-24
EditCounter: + logscale option for monthyearcharts

agoujot:
Has been requested a few times.

Implementation details:
 - linked each checkbox to both charts, because it's smoother. so like if you change one it also changes the other checkbox and both charts change scale.
 - I added for this the message `use-log-scale` in `qqq.json`, `en.json` and `fr.json`. Do tell me if there's something else I need to do to setup i18n.
 - has to start at 1, else we get a log(0) infinite loop
 - had to enforce an at least 1.5 coefficient between two ticks in `xAxis` else we'd get a lot of overlapping
 - and, and this is the part which is a bit meh, the way it works is tearing out the chart and recreating it with different options. I spent an hour or two trying to update the config dynamically, but it failed miserably every time. I don't think there's a way we can do that in chartjs 2.7. (or it isn't documented anymore... who knows.)

Bug: T177658


MusikAnimal (changes requested):
This causes my browser to crash. I used `enwiki` user `GoldenRing` as an example. That user has a long tenure, but relatively few edits. I tried on a few other accounts and had the same problem.

We appear to be using Chart.js version 2.3.0 in [Pageviews Analysis](https://pageviews.wmcloud.org), and the toggle for the logarithmic scale there seems to work fine. I don't know if any inspiration can be drawn from it (it's admittedly old and convoluted code), but here's where it happens:

* [yAxes configuration](https://github.com/MusikAnimal/pageviews/blob/72e6d45f0e9c0d2e7e48decdf61ac1095df0beb0/javascripts/shared/chart_helpers.js#L978-L995)
* [Automatically determining if logarithmic scale should be used based on the Theil index](https://github.com/MusikAnimal/pageviews/blob/72e6d45f0e9c0d2e7e48decdf61ac1095df0beb0/javascripts/shared/chart_helpers.js#L731-L768) – something we might consider as well

It does appear, however, that I also had to [destroy the chart and rebuild](https://github.com/MusikAnimal/pageviews/blob/72e6d45f0e9c0d2e7e48decdf61ac1095df0beb0/javascripts/shared/chart_helpers.js#L1000). That I suppose is fine; you don't even notice really, at least on my machine.

agoujot:
Uuhh, hang on, let me check. I must have broken it at some point in all the main merges.

agoujot:
What browser are you on?

agoujot:
I'll only be able to test this afternoon, in all likeness. That being said, it worked perfectly for me locally last time I tested, so it's likely a small bug introduced somewhere.

agoujot:
Hang on a sec. I just realised the second chart doesn't change.

agoujot:
Huurrggh. *Bangs head against desk.*
 - Year and months datasets have the same format.
 - Their labels are closely similar.
 - Can't see significant differences between the twig files.

And yet, after splitting, and even in the different subpages in different tabs:
 - year charts work smoothly.
 - month charts gulp all CPU until browser crash.

I have traced the timeout to the chartjs call proper.
(This was probably also in fact the cause of the other timeout; both checkboxes were triggering the monthcounts.)

agoujot:
*Huurrrggghhh.* After a lot of head-banging and thinking I found it three times, I finally found it. It's one of the stupidest bugs I ever saw. Until ChartJS PR [4959](https://github.com/chartjs/Chart.js/pull/4959), which was in 2.7.2 (we're in 2.7.0 :smiling_face_with_tear:), every time a log scale is given a zero, it goes infinite-loopy.

Well, anyhow. In the end the fix is `min:1,` in log scale.

Heh, what was I saying the other day about the amount of work for six bytes? This time about 5 hours. Don't mind the close-reopen, just a misclick.

MusikAnimal (approved):
Logarithmic scales can't begin at zero anyway, and there are no fractional numbers here, so starting at 1 is fine. Pageviews Analysis appears to do the same thing.

Thank you for your hard work!! :D

PR #511, opened 2025-04-24
EditCounter: make year/month chart sizing foolproof by using external legend

agoujot:
Essentially:
 - put the legend/table previously in `templates/editCounter/namespace_totals.html.twig` into `templates/editCounter/externallegend.html.twig` 
 - enable externallegend on the year/month subpages
 - make the application method for toggling work also if there's no pie chart
 - disable chartjs legends in year/month in all cases
 - change some pointer-events stuff so that the chart resize listener doesn't prevent clicks on the legend; this is needed to keep the two next to each other
 - and, finally, change height formula to `30*(rows+1)`, this time hopefully forever, and making no assumptions
 
Bug: T392604


MusikAnimal (approved):
Amazing work!! :100: 

The only issue I can find is that the responsive chart (when viewing on a phone, for example) overlaps the newly added "Use logarithmic scale" checkbox. I'm assuming that wasn't there when you wrote this PR.

I've got a long list of PRs left to review (one of those _good_ kind of problems :smile:) so I'm going to go ahead and merge. The aforementioned issue can be dealt with separately.

agoujot:
Ah, that's the issues with running 30ish parrallel PRs. Coordination of compatibility gets complicated :).

agoujot:
Should be solved with #538 .

PR #512, opened 2025-04-25
EditCounter: rm duplicate auto removals after manual removal

agoujot:
(At last, it's been reported at least four times.)

We already had code in place to do this: there was just a little bug with how `unsetAutoRemoval` worked.

It filtered rights changes to unset to keep only `$grantType === 'pending'`, but `pending` is for stuff that can't have happened yet. The past auto rights removals added by `setAutoRemovals` in fact have `$grantType === 'automatic'`. So it was just a matter of changing a word there.

As they say, a second pair of eyes...

Bug: T193838


MusikAnimal (approved):
:100: thank you!!

PR #513, opened 2025-04-25
Dark mode adaptations

agoujot:
Through css.
As the svg images are made for light mode, and can be hard to distinguish in dark mode.
As part of this, the darker and lighter colors will be swapped, but I don't think it's too much of a problem.

Bug: T327579


agoujot:
While we're at it, I also inverted the actual xtools logo. It was suffering from darkmode issues too. Compare before:
image
To after:
image

agoujot:
This was especially an issue for the home link. Before:
image
After:
image
With the the navbar being lighter, the text was barely lighter than its background.

agoujot:
Other possible issues:
 - text-muted is close to the bg
 - about the chart grids, it's a bit the other way. It looks great in dark mode, it's there and visible but not too proeminent, but in light mode... It's very very proeminent. Compare:
 image
and:
image
We're using `#333` across skins, and one of them should change.

MusikAnimal (commented):
> about the chart grids, it's a bit the other way. It looks great in dark mode, it's there and visible but not too proeminent, but in light mode... It's very very proeminent.

In my testing this is only a problem when switching between dark/light mode without a page refresh. If that is the case, I'm OK merging. Otherwise it is indeed a bit _too_ prominent, I think!

We can't use SCSS variables or the like since the colors are set in the JavaScript. I guess the only solution then is to find a suitable colour for dark mode and apply that programmatically.

agoujot:
Noting that in the end I did *not* change anything about the grid color.

PR #514, opened 2025-04-26
EditCounter: add edit size histogram

agoujot:
Essential changes are (updated):
 - change SQL query to keep everything (of the 5k, ofc)
 - add public function to get all edit sizes
 - leave the two "small edits" and "large edits" in the top-right simple "edits" table
 - replace the two big redundant pie chats with a bar chart (vertical bar chart, not horizontal bar chart, as here the space we have fits that rather). 200 px tall.
  - x axis is size intervals, 10*powers of 2. 11 bars, fits pretty nicely.
  - y axis is number of edits in that interval. additions above the x-axis, removals below the x-axis

Where before we had this:
image
Now we have this:
image

Bug: T236087


agoujot:
I for one thinks that this fits pretty nicely into where there used to be two pie charts, and that there's not really a need to move it to an external thing.

On time, getting them all doesn't take more time than counting them, so we're fine.

Also, we need to talk about the intervals. We have a real estate hard limit of 12-13 bars *max*. I chose to make equal-sized intervals, because I thought that else it can be misleading. We could go for a geometric thing like 0-50; 50-100; 100-200; 200-400; 400-800; 800-1600; 1600-3200, > 3200. That would make us 8 intervals, which is good. It would give more granularity; eg separating the typo fix of 2 from the sentence adding of 80; as opposed to the two-para addition of 600 being not that distinct from the three-para addition of 800. But that's where the risk of misreading kicks in...

(Also, about codecov, IDK how I'm supposed to test something that sends back 5k values.)

agoujot:
I've made a post at the enwp village pump to see what users think of the interval story.

agoujot:
(Was not constraining width of `basic-info-charts canvas` outside EC anymore; which is bad; so I rechanged the selector to only exclude `sizechart-canvas`.)

agoujot:
Ok so:
 - split neg, pos & zero
 - fixed a bug with the count of smallest and largest, and the average, being counted as sizes
 - switched to a log2 scale per on-wiki recommendations.
 - changed number of bars
Now looks like this:
image

agoujot:
Updated colors for accessibility:
image
(the green is colors(1) aka 55FF55, the red is colors(0) aka FF5555, and the blue is colors(4) aka 5555FF).

We've got AA contrast, and good with colorblindess.

agoujot:
With the removals below the zero line:
image

agoujot:
(Had to delete a commit with my db identifiers.)

agoujot:
With a horizontal and not vertical bar chart (which is ''not'' pushed) it gives:
image
Which is rather cramped.

assets/js/editcounter.js
```diff
@@ -306,6 +306,98 @@ xtools.editcounter.setupMonthYearChart = function (id, datasets, labels, maxTota
...
 
+/**
+ * Setup edit size histogram as a vertical bar chart
+ * from the PHP EditSizeData.
+ * @param {Object} data JSON object returned by getAllEditSizes.
+ * @param {Array} colors CSS colors for additions, removals, and same-size, in that order.
+ * @param {Array} barLabels i18n'd bar labels for additions, removals and same-size, in that order.
+ */
+xtools.editcounter.setupSizeHistogram = function (data, colors, barLabels) {
+    let bars = 11;
```
MusikAnimal, line 317: Should be a `const`?

src/Model/EditCounter.php
```diff
@@ -1069,4 +1069,18 @@ public function averageEditSize(): float
...
     }
+
+    /**
+     * Get the sizes of the user's past 5000 edits.
+     * return int[] Array of sizes in bytes.
+     */
+    public function getAllEditSizes(): array
+    {
+        $editSizeData = $this->getEditSizeData();
+        if (isset($editSizeData)) {
```
MusikAnimal: `isset()` is used in the other methods to check for the existence of an array key, because PHP would otherwise throw a warning about undefined array index (though nowadays we should be using the [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op)). Here we don't need to check anything since we want the whole array, and it's guaranteed to be an array thanks to the `array` return type and that we use [strict mode](https://stackoverflow.com/questions/48723637/what-do-strict-types-do-in-php).

So, this method could be simplified to `return $this->getEditSizeData()`, effectively an alias. Perhaps instead the callers should use `getEditSizeData()` directly.

templates/editCounter/general_stats.html.twig
```diff
@@ -514,34 +514,20 @@
...
-                    percentage: 100 - ((ec.countLargeEdits / ec.countLast5000) * 100)
-                },
-                {
-                    label: '≥ ' ~ 1000|num_format ~ ' ' ~ msg('num-bytes', [1000]) ~ '*',
-                    value: ec.countLargeEdits,
-                    percentage: ((ec.countLargeEdits / ec.countLast5000) * 100)
-                }]
-            )
-        }}
+        <div style="float:left">{{ msg('num-edits-size-interval') }}*
```
MusikAnimal: This gets flipped in RTL. You can use the [Bootstrap](https://getbootstrap.com/docs/3.4/css/#helper-classes-floats) `pull-left` class and that will get flipped by the RTL stylesheet.

We should instead be using [CSSJanus](https://www.npmjs.com/package/cssjanus), but that might necessitate [T392531](https://phabricator.wikimedia.org/T392531). Anyway, not for this PR.

templates/editCounter/general_stats.html.twig
```diff
@@ -514,34 +514,20 @@
...
+        <div style="position:relative;height:200px;min-width:650px">
```
MusikAnimal: I suggest applying a class and using stylesheets.

MusikAnimal (changes requested):
Amazing stuff! I think it looks great.

A couple of issues and suggestions, nothing too big.

agoujot:
Think I've addressed all comments.

MusikAnimal:
I don't know what I did to mess up the PR… I just merged manually on my local. Commit is 7d1f26d18e50279725fdc69ec4f04c72454a9319 with attribution intact

agoujot:
Well, whatever works is good.

PR #515, opened 2025-04-26
EditCounter: simplify greatly dataset creation for charts

agoujot:
We've used twig loops to create JS code, and not well, which resulted in absurdly large JS for what it was doing (for my monthcounts, 180k).

I made some simplifications using twig fitlers and json_encode().

Resulting JS code is at least 4x smaller, up to 14x smaller for month charts.

Bug: T392731


MusikAnimal (approved):
:100: Thanks!

PR #518, opened 2025-04-29
Fix html entity escaping in summaries

agoujot:
Essentially:
 - remove htmlspecialchars all over the place in sections (some parts of a string could be escaped up to three times!)
 - decode twice for the section URL - once because we add one, and once because of possible user entities (MW decodes when determinating section IDs)

Bug: T261904


src/Model/Edit.php
```diff
@@ -345,7 +345,7 @@ public static function wikifyString(
         ?Page $page = null,
         bool $useUnnormalizedPageTitle = false
     ): string {
-        $summary = htmlspecialchars(html_entity_decode($summary), ENT_NOQUOTES);
+        $summary = htmlspecialchars($summary, ENT_NOQUOTES);
```
MusikAnimal: What happens when `$summary` is `<script>alert('I steal yo cookies!');</script>`? I think the `msg()` helpers in the templates will ultimately escape the HTML, but best to be sure :)

agoujot: Per [the doc](https://www.php.net/manual/en/function.htmlspecialchars.php), that's what `htmlspecialchars` takes care of: eg for that summary gives `&lt;script&gt;alert(&#039;I steal yo cookies!&#039;);&lt;/script&gt;`.

If the summary entered is `& is not &amp; also <script>alert('I steal yo cookies!');</script>`, current code shows `& is not & also <script>alert('I steal yo cookies!');</script>`; while the proposed code shows exactly what was entered.

I'll note that this (conflating characters and their entities) is also current MW behaviour. That seems like a bug more than a feature to me, but if you want we can readd the `html_entity_decode`.

agoujot: After thinking some more about this I decided to readd it to be consistent with MW behaviour.

PR #521, opened 2025-04-30
TopEdits and PagesCreated: add list of ten top wikiprojects

agoujot:
DEWISOTT. (Figuring out the queries took a damn long time.)

On performance: same for topedits (just added an instant subquery), and we're at about +1 second for a user with 20K edits for pages. (Had to make new query.) If you've got an idea of how to integrate this into another query, feel free to. Just have bashed my head at SQL for too long today to be in a state to figure it out. Maybe will tomorrow. Didn't do for editcounter, as I don't know where we'd put it; will possibly also do tomorrow.

Bug: T344464


agoujot:
Hang on, currently not existing in the wikitext.

agoujot:
Maybe the 10 should be an env param.

agoujot:
Normally no memory issues: limit 10 on one side, only adding a column on the other.

agoujot:
(CI passes locally.) Note: enwiki has just added to about all assessed mainspace pages a supplementary assessment with project named "Project-independent quality assessment"; which means it'll top every list, and essentially give the number of pages. I'm not sure how we're supposed to deal with that: it's a hack with no explicit support in the extension, and we shouldn't hard-code stuff.

src/Model/Pages.php
```diff
@@ -301,6 +301,37 @@ public function getAssessmentCounts(): array
...
+                $this->project,
+                $this->user,
+                $this->namespace,
+                $this->redirects
+            );
+        } else {
+            $counts = [];
+            foreach ($this->pages as $nsPages) {
+                foreach ($nsPages as $page) {
+                    if (!isset($counts[$page['assessment']['project']])) {
```
MusikAnimal: I'm getting `Warning: Undefined array key "project"` for this line for any user, i.e. http://localhost:8000/pages/en.wikipedia.org/MusikAnimal/0

`isset()` won't check for existence of nested arrays.

The question of course being, how'd we get an assessment without a project?

src/Model/TopEdits.php
```diff
@@ -155,6 +155,34 @@ public function getNumTopEdits(): int
...
+    {
+        $projectTotals = [];
+        // List of pages for this namespace
+        $rows = $this->topEdits[$ns];
+        foreach ($rows as $row) {
+            $num = $row["count"];
+            // May be null or nonexistent for assessment-less pages
+            $titles = $row["pap_project_title"] ?? "{}";
+            // Had to use json to pass multiple values in SQL select
+            foreach (JSON_decode($titles) as $projectName) {
```
MusikAnimal:
```suggestion
            foreach (json_decode($titles) as $projectName) {
```
that's the [documented](https://www.php.net/manual/en/function.json-decode.php) form. Maybe we should add a linter rule for this.

MusikAnimal (commented):
Will finish the review tomorrow but I thought I'd leave this early feedback.

src/Model/TopEdits.php
agoujot: Done

src/Model/Pages.php
agoujot: I'd just forgotten to fetch the projects in getPagesCreated. Probably during testing I only took users where we took the first branch of the if (using getWikiprojectCounts, when some pages weren't analysed in getPagesCreated). Now done.

templates/topedits/result_namespace.html.twig
```diff
@@ -23,69 +23,105 @@
...
+                            <td colspan={{ showPageAssessment ? 4 : 3 }}>
+                                <a href="{{ path('TopEditsResultNamespace', {project:project.domain, username:user.usernameIdent, namespace:ns}) }}">
+                                    {{ msg('more') }}&hellip;
+                                </a>
+                            </td>
+                        </tr>
+                    {% endif %}
+                </tbody>
+            </table>
+            {% if showPageAssessment %}
```
MusikAnimal: I suggest only showing this on the mainspaces subrequest for performance. The "all namespaces" TopEdits is already a bit slow. What do you think?

templates/topedits/result_namespace.html.twig
```diff
@@ -23,69 +23,105 @@
...
+                                    {{ msg('count') }}
+                                    <span class="glyphicon glyphicon-sort"></span>
+                                </span>
+                            </th>
+                        </tr></thead>
+                        <tbody>
+                            {% for wikiproject, count in totals %}
+                                <tr>
+                                    <td class="sort-entry--wikiproject" data-value="{{ wikiproject }}">
+                                        {{ wikiproject }}
```
MusikAnimal: Ideally we'd link to these [like we do in PageInfo](https://github.com/x-tools/xtools/blob/681e2c219b0e927501beaecf6cfa18748696bc7d/templates/pageInfo/result.html.twig#L960)

src/Model/Pages.php
```diff
@@ -301,6 +301,39 @@ public function getAssessmentCounts(): array
...
+            $counts = [];
+            foreach ($this->pages as $nsPages) {
+                foreach ($nsPages as $page) {
+                    foreach ($page['assessment']['projects'] as $project) {
+                        if (!isset($counts[$project])) {
+                            $counts[$project] = 1;
+                        } else {
+                            $counts[$project]++;
+                        }
+                    }
```
MusikAnimal, line 326: Maybe something like this would be cleaner:
```
foreach ($page['assessment']['projects'] as $project) {
    $counts[$project] ??= 0;
    $counts[$project]++;
}
```

MusikAnimal (commented):
In addition to comments, I think the older styles for the caption looked nicer. After this PR, there is a vertical border in the <thead> that doesn't connect to anything.

templates/topedits/result_namespace.html.twig
agoujot: Why not.

agoujot: Why not. As in /all the limit per ns is very low, we'd nearly always have to make the separate subquery, too.

agoujot:
The caption thing was an unintended effect of a `display` change. fixed by wrapping in divs.

templates/topedits/result_namespace.html.twig
agoujot: I didn't know we had that set up, that's really cool. I'll look into doing this in TE too.

agoujot: Done! Thankfully, PageAssessments is exposed through Project.

agoujot:
And, of course, "Project-independent assessment" bites again: linking to a page for this not-a-wikiproject. IMO, that was a bad idea. If it's projectless, leaving project empty would have made far more sense and been much less language-dependent. On top of that, task forces and work groups sometimes don't have a page. Perhaps we should ignore them? There's an easy way to detect them, at any rate (those with a non-null parent project id).

src/Model/Pages.php
agoujot: Sure would be, and also could apply in a few other places in my code I think; Didn't think of it this way.

agoujot: Done.

src/Repository/PagesRepository.php
```diff
@@ -350,6 +358,71 @@ public function getAssessmentCounts(
...
+
+        // index => [name, count]
+        $result = $resultQuery->fetchAllNumeric();
+        // convert that to: name => count
+        $totals = [];
+        foreach ($result as [$name, $count]) {
+            $totals[$name] = $count;
+        }
+        // sort by count decreasing
+        arsort($totals);
```
MusikAnimal:
```suggestion
        $totals = $this->executeQuery($sql, $project, $user, $namespace)
            ->fetchAllAssociative();
```

templates/pages/result.html.twig
```diff
@@ -161,41 +161,74 @@
...
+                                            {{ wikiproject }}
+                                        </td>
+                                        <td class="sort-entry--count" data-value="{{ count }}">
+                                            {{ count|num_format }}
+                                        </td>
+                                    </tr>
+                                {% endfor %}
+                            </tbody>
+                        </table></div>
+                        {% endif %}
```
MusikAnimal, line 197: Something is off here with the alignment. It looks like the `<div class="side-by-side">` should be outside this `if` block?

templates/pages/result.html.twig
```diff
@@ -161,41 +161,74 @@
...
-                                    </td>
-                                    <td class="sort-entry--count" data-value="{{ count }}">
-                                        {{ count|num_format }}
-                                    </td>
-                                </tr>
-                            {% endfor %}
-                        </tbody>
-                    </table>
+                    {% set totals = pages.wikiprojectCounts %}
+                    {% if totals|keys|length > 0 %}
```
MusikAnimal: This or something inside this block is throwing a `Warning: Array to string conversion` when loading results for `all` namespaces

MusikAnimal (changes requested):
> If it's projectless, leaving project empty would have made far more sense and been much less language-dependent. On top of that, task forces and work groups sometimes don't have a page. Perhaps we should ignore them?

Extension:PageAssessments currently [requires a project name](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/PageAssessments/+/9437d24d032b418223bee2e255bd861bdd0dc692/src/PageAssessmentsDAO.php#69). I guess there's a use-case here to upstream the concept of general, project-agnostic assessments. But then again, I guess it's easy to just give it a name. Anyway, you're right about making assumptions when linking. Let's just forget the links.

src/Repository/PagesRepository.php
agoujot: Fair enough.

templates/pages/result.html.twig
agoujot: Indeed.

templates/pages/result.html.twig
agoujot: Due to a different structure depending on whether we got it from the repo or from the already-compiled data. Changed the format of one to go with the other.

templates/topedits/result_namespace.html.twig
```diff
@@ -23,69 +23,105 @@
...
+                                    {{ msg('count') }}
+                                    <span class="glyphicon glyphicon-sort"></span>
+                                </span>
+                            </th>
+                        </tr></thead>
+                        <tbody>
+                            {% for wikiproject, count in totals %}
+                                <tr>
+                                    <td class="sort-entry--wikiproject" data-value="{{ wikiproject }}">
+                                        {{ wiki.pageLinkRaw(te.wikiprojectPrefix ~ wikiproject, project, wikiproject|ucfirst) }}
```
MusikAnimal: Let's just remove the link since we've concluded we can't blindly link (or don't link for subprojects, as you suggested). We can deal with the project-independent assessments in a follow-up

src/Repository/TopEditsRepository.php
```diff
@@ -122,6 +122,16 @@ public function getTopEditsNamespace(
...
+        $paProjectsTable = $project->getTableName('page_assessments_projects');
+        $paProjectsSelect = $hasPageAssessments
+            ? ", (
+                    SELECT JSON_ARRAYAGG(pap_project_title)
+                    FROM $paTable
+                    JOIN $paProjectsTable
+                    ON pa_project_id = pap_project_id
+                    WHERE pa_page_id = page_id
+                ) AS pap_project_title"
+            : '';
```
MusikAnimal, line 134: This query is intended to match up with the current page of results (and all the edits it encompasses), so it is not suitable for a "Top WikiProjects" table. If you change to another page, you get different totals, which I think is confusing. The existing subquery on `page_assessments` is for the "Assessment" column.

I do really like the idea of putting this info in TopEdits, though… it feels like the right place. Seeing your history by WikiProject is also really useful, I think. It's nice we have in Pages Created but that is only a fraction of the editing activity.

It would be a lot of work, but I would recommend one of the following:
* If we want to keep "Top WikiProjects" in TopEdits
  * Add a "Show" option to the TopEdits form that allows users to opt-in to showing additional statistics. This would be similar to the "Show" options seen on the Edit Counter form.
  * We might as well offer "Top Assessments" as well, so there'd be two options, both off by default (for performance)
  * Put the tables of results in a different section, let's call it "Summary". That section could also include the total number of pages (currently shown inside the namespace section).
  * Add charts or whatever we want to make it look pretty :sparkles: 
* Move to a different tool
  * We could call it "Assessments"
  * Would contain the same info as described above, but simply put in a different tool for performance reasons.

For now, I would suggest removing "Top WikiProjects" from Top Edits, just to get this PR moving along. I'm content with the implementation in Pages Created.

MusikAnimal (commented):
:+1: on Pages Created.

templates/topedits/result_namespace.html.twig
agoujot: Indeed.

src/Repository/TopEditsRepository.php
agoujot: Indeed, hadn't taken pagination into account.
Feel free to slap me; but couldn't we just do like we do in Pages? As in, if all results fit on the current page, then return something based on that; else, do a separate query. Too slow, I suppose?

MusikAnimal: Pages Created only looks at revisions with no parent, so it in general will be considerably quicker than checking every page edited by a user. Similarly, TopEdits requires pagination more often than Pages Created, so I imagine many times if not most of the time, it'd have to take the slower route of running a separate query. It may still run reasonably quick though; I just had gut instincts because TopEdits already seems slow to me.

If you want to code it up, feel free and I will help with testing. If it turns out to be too slow, we can save your work for a separate PR, going with either route mentioned above.

agoujot: Working on it. Noticed bug while doing this: was not passing start and end to the PagesRepository method.

agoujot: Done! Actually, it's quite a surprise to me, but this is fast: only takes ~2-3 seconds for https://xtools.wmcloud.org/topedits/en.wikipedia.org/Graham87/0 (123k pages in ns0; 237k edits in ns0).

MusikAnimal: Great test subject! Graham has had quite an influence on XTools, in particular screen-reader compatibility :) His accoun,t given the tenure and prolificacy, is great for testing!

On my end, this query took ~5.8 seconds. I don't know if I mentioned it before, but simply re-running fresh queries won't prove much necessarily, as the MariaDB buffer pool may have the results already in memory (or something to that effect). I usually wait ~10 minutes to do run time tests, but there's no definitive. The `EXPLAIN` results are our best bet for measuring performance, and it looks concerning for this query and Graham's example. Filesorts can be OK but here that was the first query plan.

MusikAnimal (approved):
I mentioned the poor query plan, but this may _still_ be fine… MariaDB/MySQL is very smart and, after all, the entire point of XTools is "give me all the data!". Everyone knows they have to wait.

I'm happy to just merge to get this moved along. I imagine you're exhausted, and this is a clear improvement that many will enjoy! :smile: Let's get it out, and we'll refine as necessary. I'm going to remove XFF from emails and get you subscribed, then you'll at least see which queries are the slowest.

PR #522, opened 2025-05-03
Remove obsolete i18n messages

agoujot:
(Noticed with #510.)

I'm pretty sure that these messages are not used anymore and can be removed to take the load off of the translators:
```
accounts
added
all-namespaces
average-change-per-page
considered-usernames
edits-in-category
events
file
invalid-date
invalid-type
minor
neutral
next
no-exist
num-blocks
num-users
oppose
others
pageviews-pages-created
prev
result
result-by-name
result-by-time
size-over-time
support
tagged
text
text-to-search
tool-about
type
user-is-not-admin
vote-msg
wiki
```
There are 33, i.e. ~7% of all messages.

Please do double-check, especially for the short ones; I did my best to filter but some of these might be used by the php code in a  way I didn't notice.

Bug: T393120


agoujot:
Readded `unique-pages-edited` per use in #574.

i18n/en.json
```diff
@@ -438,6 +406,8 @@
 	"tool-authorship-desc": "Authorship attribution, measured by character count, excluding spaces",
 	"tool-autoedits": "Automated Edits",
 	"tool-autoedits-desc": "Analysis of semi-automated and non-automated contributions",
+	"tool-bash": "Random Quote",
+	"tool-bash-desc": "Get quotes from the IRC Quote Database",
```
MusikAnimal: The tool is called Quote, so we should stick with that, I think, and instead fix the `xtPageTitle` and `xtSubtitle` in QuoteController::indexAction()

i18n/en.json
```diff
@@ -284,7 +269,6 @@
 	"num-edits-size-interval": "Number of edits in that size change interval, in bytes",
 	"num-hours": "{{PLURAL:$1|hour|hours}}",
 	"num-minutes": "{{PLURAL:$1|minute|minutes}}",
-	"num-months": "{{PLURAL:$1|month|months}}",
```
MusikAnimal, line 287: Let's keep this along with `num-weeks` and `num-years` for https://phabricator.wikimedia.org/T406182

src/.env.local
```diff
@@ -0,0 +1,87 @@
...
+
+# Oauth consumer credentials, from https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration
+OAUTH_KEY=
+OAUTH_SECRET=
+
+###> symfony/mailer-bundle ###
+MAILER_DSN=null://localhost
+MAILER_FROM_EMAIL=tools.xtools@toolforge.org
+MAILER_TO_EMAIL=tools.xtools@toolforge.org
+###< symfony/mailer-bundle ###
```
MusikAnimal: Added by accident I presume? I'm relieved there are no real credentials in here!

MusikAnimal (changes requested):
Few minor things, otherwise looks great! Thank you for doing this tedious work :pray:

i18n/en.json
agoujot: Ok!

i18n/en.json
agoujot: Fair enough.

PR #523, opened 2025-05-03
PageInfo: cache getBotData

agoujot:
I was on my way to add caching to it all, but actually only one query was missing it: getBotData.

That's pretty much it :). (oh, and had to tweak use a bit to return array not resultstatement (which should be fine as we iterate over it all anyhow.)

Bug: T208543


agoujot:
Note on memory: caching has necessitated using an array and not resultstatement. This perhaps puts more stuff in memory. However, both of the two subqueries unioned here use `LIMIT`ed to `MAX_PAGE_REVISIONS`, which is 50k in current setup, and so ought to be fine. On top of that, what the queries return are counts after `GROUP BY`, so very likelily the result contains less rows than the sub-subqueries (if that word makes sense).

agoujot:
Update: tried with WP:AIV, which a) is above 50k and b) is one of the most bot-edited pages of enwp; worked without complaints. I'm merging this.

agoujot:
Leaving uncached the links query from Repository/PageRepository, that one's very fast.

PR #524, opened 2025-05-05
EditCounter: replace DESC by minus in log_latest ORDER BY

agoujot:
For some reason it's much faster. Details at the phab task.

Bug: T387210


agoujot:
Now we know why it works! :partying_face: In a nutshell, `userindex` makes `log_timestamp` indexing a very bad idea (plus, MySQL doesn't know about backwards index scan). See https://phabricator.wikimedia.org/T387210#10802174 for details.

We might as well put `(0+log_timesamp) DESC` if we want.

(it's incredible the amount of research that can go into a six-byte change. :))

PR #525, opened 2025-05-08
Add link loading timers

agoujot:
That's pretty much it. Also does for this what #517 does for form timers.

Bug: T177902


agoujot:
**Don't merge:** mixing this with the other one was a bit messy; and apparently cdnjs's having trouble so I can't test properly yet.

MusikAnimal (changes requested):
This is nice! A few suggestions:

* Don't add the timer for links with `target="_blank"`, such as the "Edit Summaries" and "Automated Edits" links in the "_Edits (live)_" section of the PageInfo tool.
* Add a CSS class, say `.link-loading`, that will gray it out and a `pointer-events: none` to prevent re-clicking.

agoujot:
Done! (Given all that color entails accessibility-wise, I preferred to go with italics instead of graying out.)

PR #527, opened 2025-05-09
Adapt better to invalid language codes

agoujot:
Namely, fall back to `en` earlier (in `getIntuitionLang`) if Intuition says this isn't a valid language code, and flash to the user that it isn't. Because `$.i18n` (used in charts and a few other places) crashes when fed an invalid language code.

Bug: T393736


agoujot:
Oh, and if you know how to make error notices from the i18n helper it would be clearer to end users what exactly happened. I couldn't find something exposed to this file to do this.

agoujot:
#571 did this better; no need to do this; closing.

PR #528, opened 2025-05-09
SimpleEditCounter: add creation count

agoujot:
Have checked performance: after merging with the regular edits subquery, shouldn't be any slower.

Bug: T229581


agoujot:
FYI, the countsOnly option in pages is #550.

agoujot:
Hang on: forgot to filter by namespace when applicable.

PR #529, opened 2025-05-10
AdminStats: add footnote for (re)protect and (re)block

agoujot:
Specifying they include both the normal action and the "re-" action.

Bug: T354024


agoujot:
Hang on: forgot that protect & block aren't always included. Until I correct this UI issue, do *not* merge.

MusikAnimal:
> Hang on: forgot that protect & block aren't always included. Until I correct this UI issue, do _not_ merge.

Set as a draft to prevent merging. I didn't see your comment until it was too late, hehe

PR #531, opened 2025-05-25
GlobalContributions: Highlight pending edits (FlaggedRevs extension)

agoujot:
Changes are roughly:
 - add method to project to get if a *given* project has FlaggedRevs on. for this also tweak getinstalledextensions to use projectrepository's capacity to get *any* project's (not only $this's) extensions
 - in globalcontribs' getrevisions, for each project, if it has flaggedrevs, outer join on flaggedpages and select whether it's pending( this check is actually ugly due to how flaggedrevs works)
 - add the pending property to edit, and an exposed method to get it, and set it using attrs' is_pending
 - add i18n for pending msg
 - add styling for pending (make it stand out just a bit less than reverted)
 - highlight and label in twig

Bug: T322466


agoujot:
Oh hang on, this may have a problem: getting the extensions of 1 project is 1 api call. From quick testing, these api calls each take 0.3-08 seconds. Let's say average 0.5. That means, for someone that has edited 50 projects, or 5% of all projects, 25 seconds of waiting time. That's not good.

agoujot:
Similarly to the fix for #568 , normally we should be able to check having or not flaggedrevs simply with `information_schema.tables`.

PR #532, opened 2025-05-25
EditCounter: + approx autoedits count using tags

agoujot:
Added just above the small/large edits.

Changes:
 - add `auto-edits-tagged` message for this
 - add a getTags method to autoeditshelper to just get the tag list
 - inject autoeditshelper into editcountercontroller and editcounter
 - get the revision tags in getEditSizeData in editcounterrepository
 - add an editcounter method to use getEditSizeData and getTags to get how many edits are automated in the last 5000
 - add a row in the html table in general_stats.html.twig

Bug: T342402


agoujot:
I'm uncertain about the label: it really should be short, but it should be clearly said that this is an approximation. Meanwhile, I settled for `Tagged (semi-)automated edits`, but it's still longer than all the other labels.

agoujot:
And I suppose the page join I removed was for T325492? I spose I might as well readd it.

templates/editCounter/general_stats.wikitext.twig
```diff
@@ -90,6 +90,10 @@
 | {% verbatim %}{{FORMATNUM:{% endverbatim %}{{ ec.countMinorRevisions }}}}{% if ec.countLiveRevisions %} ({{ ((ec.countMinorRevisions / ec.countLiveRevisions) * 100)|percent_format }})
 {% endif %}
 |-
+| {{ msg('auto-edits-tagged') }}
```
MusikAnimal:
```suggestion
| {{ msg('auto-edits-tagged') }}*
```

MusikAnimal (commented):
One tiny little requested change to the wikitext output. Otherwise this looks good and I'm happy to approve.

It'd be _nice_, but not required, to have a unit test to reinforce the `EditCounter::countAutoEdits()` method, and perhaps also `AutomatedEditsHelper::getTags()`. For the former, you could stub `EditCounterRepository::getEditSizeData()` to return the data structure we expect, and have it include a tag or two that we don't consider automated. You'll also need to stub `AutomatedEditsHelper::getTags()` to return tags we do consider automated. We only need a few examples: say, `AWB`, `contenttranslation`, and `mobile edit`. AWB counts as automated. `contentranslation` would also, except that it has `contribs: true` which I think means we don't want to count it? Then `mobile edit` is one we wouldn't count.

Let me know if you need help!

agoujot:
@MusikAnimal : if you don't mind, I'd rather make that test after we merge this or #551 , because 551 already has the provider &c set up to do this. If we merge 551 then adapting this (relying on that) will be easy; if we merge this first I can add the tests to 551.

MusikAnimal (approved):

PR #533, opened 2025-05-26
AutoEdits: actually use `tag_excludes` correctly

agoujot:
To be precise,
 - actually exclude excluded tags in `autoedits-contributions` when giving a `?tool`
 - same for the individual tool count in the table. I left the footnote, because these are the overlaps `tag_excludes` describes; it's still imperfect.

Implementation details (all changes to `src/Repository/AutoEditsRepository.php`):
 - get `$tagExcludesIds` in `getToolRegexAndTags`; return as third array element
 - use it in `getAutomatedEdits` (no need for `getNonAutomatedEdits`, as it doesn't care how many tools an edit uses)
 - fix `getTagsExclusionsSql` (for the table):
  - why `&& 1 === count($tagIds)`? some tools have multiple tags IIRC
  - we're `LEFT OUTER JOIN`ing `change_tag` on `revision`, so each row contains only one tag of the revision; therefore using `AND ct_tag_id NOT IN` was never going to work; have to go the long way round with `AND NOT EXISTS (SELECT 1 FROM ... WHERE ct_tag_id IN`

Bug: T382773


bunnypranav:
@MusikAnimal Any chance you could review and push this? Thanks!

MusikAnimal (approved):
Sorry for the long wait! 3 months… :facepalm:

Everything works great in my testing, merging!!

bunnypranav:
Thanks! :)

PR #534, opened 2025-05-27
EditCounter: integrate flaggedrevs in protection counts

agoujot:
Of course, FlaggedRevs' documentation is not much to go on, but from logs and quarry &c it looks like there are four actions of type `stable`:
 - `config`: apparently sets pending changes, so I bundled it in protection
 - `modify`: obviously goes with reprotection
 - `move_stable`: bundled it too with reprotection
 - `reset`: I'm pretty sure this is unprotection.

Bug: T284876

### DIFF
--- a/src/Model/EditCounter.php
+++ b/src/Model/EditCounter.php
@@ -459,7 +459,8 @@ class EditCounter extends Model
     public function countPagesProtected(): int
     {
         $logCounts = $this->getLogCounts();
-        return $logCounts['protect-protect'] ?? 0;
+        return $logCounts['protect-protect']
+            + $logCounts['stable-config'];
     }
 
     /**
@@ -469,7 +470,9 @@ class EditCounter extends Model
     public function countPagesReprotected(): int
     {
         $logCounts = $this->getLogCounts();
-        return $logCounts['protect-modify'] ?? 0;
+        return $logCounts['protect-modify']
+            + $logCounts['stable-modify']
+            + $logCounts['stable-move_stable'];
     }
 
     /**
@@ -479,7 +482,8 @@ class EditCounter extends Model
     public function countPagesUnprotected(): int
     {
         $logCounts = $this->getLogCounts();
-        return $logCounts['protect-unprotect'] ?? 0;
+        return $logCounts['protect-unprotect']
+            + $logCounts['stable-reset'];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Model/EditCounter.php
```diff
@@ -469,7 +470,9 @@ public function countPagesProtected(): int
     public function countPagesReprotected(): int
     {
         $logCounts = $this->getLogCounts();
-        return $logCounts['protect-modify'] ?? 0;
+        return ($logCounts['protect-modify'] ?? 0)
+            + ($logCounts['stable-modify'] ?? 0)
+            + ($logCounts['stable-move_stable'] ?? 0);
```
MusikAnimal: `move_stable` I think is just logged when the PC protection is moved to another title during a page move. Thus, we shouldn't count those. We don't count `protect-move_prot` for the same reason.

MusikAnimal (changes requested):
I had to merge in `main` for this to work, FYI.

Two problems:
* We shouldn't count moving a protection as a protection modification (see comments)
* The UI will make the counts link to Special:Log for the specific type and action. In this case, we're counting from two different logs, so there's no single place to link to :( Maybe it doesn't really matter, but I imagine someone will eventually point out the discrepancy. What do you think? Maybe FlaggedRevs counts _should_ be separate from the (re-)protect counts?

src/Model/EditCounter.php
agoujot: Ok, good to know.

agoujot:
> The UI will make the counts link to Special:Log for the specific type and action. In this case, we're counting from two different logs, so there's no single place to link to :( Maybe it doesn't really matter, but I imagine someone will eventually point out the discrepancy. What do you think? Maybe FlaggedRevs counts _should_ be separate from the (re-)protect counts?

We've been doing it like this in adminstats for a while and no one seems to mind, with the only complaint being that we don't do the same in EC, so I gather it'll be fine.

src/Repository/EditCounterRepository.php
```diff
@@ -211,6 +210,10 @@ public function getLogCounts(Project $project, User $user): array
             'pagetriage-curation-reviewed',
             'pagetriage-curation-reviewed-redirect',
             'pagetriage-curation-reviewed-article',
+            'stable-config',
+            'stable-reset',
+            'stable-modify',
+            'stable-move_stable',
```
MusikAnimal: No longer used

PR #535, opened 2025-05-27
AdminStats: add a "totals" row at the top

agoujot:
Containing the column sums, and using `<i>{{ msg('totals') }}</i>` instead of a username. Not much more to say.

Bug: T395311


src/Model/AdminStats.php
```diff
@@ -229,6 +229,27 @@ private function groupStatsByUsername(array $data): array
...
 
+    /**
+     * Get the "totals" row.
+     * @return array containing as keys the counts.
+     */
+    public function getTotalsRow(): array
+    {
+        $totalsRow = [];
+        foreach ($this->adminStats as $object) {
+            foreach ($object as $key => $value) {
```
MusikAnimal: `$object` sounds like the [object type](https://www.php.net/manual/en/language.types.object.php) (where members are accessed ala `$object->key`), but really this is an associative array. Something generic like `$row` or `$data` is fine.

_Nitpicky_: here we know what the keys and values represent, so `$key` could be `$action` and `$value` could be `$count`.

src/Model/AdminStats.php
```diff
@@ -229,6 +229,27 @@ private function groupStatsByUsername(array $data): array
...
+                if ('username' !== $key && 'user-groups' !== $key) {
+                    if (isset($totalsRow[$key])) {
+                        $totalsRow[$key] += $value;
+                    } else {
+                        $totalsRow[$key] = $value;
+                    }
+                }
+            }
```
MusikAnimal: The full nitpick:

```suggestion
        foreach ($this->adminStats as $data) {
            foreach ($data as $action => $count) {
                if ('username' === $action || 'user-groups' === $action) {
                    continue;
                }
                $totalsRow[$action] = ($totalsRow[$action] ?? 0) + $count;
            }
        }
```

or use a null coalescence assignment to set the counter:

```
$totalsRow[$action] ??= 0;
$totalsRow[$action] += $count;
```

templates/adminStats/result.html.twig
```diff
@@ -93,6 +93,19 @@
...
+                        <td><i>{{ msg('total') }}</i></td>
+                        <td></td>
+                        <td></td>
+                        {% if as.actions|length > 1 %}
+                            <td class="sort-entry--total" data-value="{{ totals.total }}">{{ totals.total|num_format }}</td>
+                        {% endif %}
+                        {% for action in as.actions %}
+                            <td class="sort-entry--{{ action }}" data-value="{{ totals[action] }}">{{ totals[action]|num_format }}</td>
+                        {% endfor %}
+                    </tr>
```
MusikAnimal: I think semantically this should go in the `<thead>`, which also visually separates it from the actual data set.

MusikAnimal (changes requested):

src/Model/AdminStats.php
MusikAnimal: Also, it seems like it should be straightforward to add a test case for this method to `AdminStatsTest`, should you feel compelled :)

templates/adminStats/result.html.twig
MusikAnimal: In my opinion keeping as `<td>` looks nice; No need to use `<th>` just because you're in a `<thead>`. But either is fine.

src/Model/AdminStats.php
agoujot: My bad; when I see "object" in my head I always think about JSON object literals, which are associative arrays. Why can't different languages have the same terminology, eh?

src/Model/AdminStats.php
agoujot: BTW, on null coalescing: IIRC there's quite a lot of the older code that goes the verbose "if not there then 1 else += 1" way.

Added a test. (The issue with statistics is that it's often barely testable. LIke, 5k values? So in the end most of the time I don't even check.)

templates/adminStats/result.html.twig
agoujot: Ok, done. (Had to do a dark mode fix on the way.)

agoujot:
As a side note, is there a more semantically appropriate element than `<i>` for this? Or perhaps the fact that we're in `<thead>` already carries all there is.

MusikAnimal (approved):
> As a side note, is there a more semantically appropriate element than `<i>` for this? Or perhaps the fact that we're in `<thead>` already carries all there is.

Yeah, being in the `<thead>`, I don't think the "Total" cell needs any emphasis. Feel free to remove the `<i>` if you want then self-merge, or merge as-is :)

agoujot:
How come html tags not escaped in PR messages? Tsk. Took me a few tries to parse sense out of your message. (I took the liberty to backtick those tags).

Thing is, I think it deserves italic because I had no other place to put it than the "username" column, and I don't want to get it taken for a username. Just not having a number, not being linked, having 0 listed usergroups, and no "ec • guc • log" links might not be enough to differentiate that row from a normal row. Our `<thead>` styling is not very prominent.
Self-merging as-is.

PR #536, opened 2025-05-28
EditCounter: localise numbers in time card and meta charts

agoujot:
Was already done for all other charts we use.

Bug: T218267


agoujot:
Oh, and also did it for meta while we're at it.

PR #538, opened 2025-05-29
EditCounter: fix interaction between external legend and log checkbox with float:left

agoujot:
The chart was getting pushed down by the legend but probably because of pos:relative or some other styling the checkbox stayed where it was, which made it overlap. Solution: use flexbox instead of floating stuff.

Follow-up to  #511 / #507 .


templates/editCounter/monthcounts.html.twig
```diff
@@ -39,7 +39,8 @@
 
     {# Set height of chart based on how many months (horizontal rows) are being reported. #}
     {# the pointer events is to not capture legend clicks #}
-    <div style="position:relative; pointer-events:none; height:{{ 30 * (ec.monthCounts.monthLabels|length + 1) }}px">
+    {# float-left to avoid issues with the external legend and the checkbox #}
+    <div style="float:left;position:relative; pointer-events:none; height:{{ 30 * (ec.monthCounts.monthLabels|length + 1) }}px">
```
MusikAnimal: How does this look in RTL? We might need a CSS class for these canvas wrapper `<div>`s, and put all but the `height` rule there. Then you can add the `float:right` it in `_rtl.scss`

agoujot: Setting out to see "How does this look in RTL?", it appears we've also got other problems. Namely,
 1. float compresses the chart and makes it really, well, not wide.
 2. for some reason, it appears that logarithmic mode is completely broken in RTL languages. In Bengali (LTR):
image
 In Arabic (RTL):
image
Some very weird things are happening.

agoujot: Oh, and it appears that resizing is also broken for RTL languages.

agoujot: I've got a fix for the original issue, float messups with directionality, and float messups with size: just put the external legend after the rest. But not for the last ones (log and resizing broken).

agoujot: Ah, I think I know why log breaks: my check to enforce 1.5* distance between ticks assumes a single direction. EDIT: was not just that; log still *really* doesn't like reverse.

agoujot: Welp, just disabled reverse if logarithmic *shrug*.

MusikAnimal (approved):
So it seems like the main goal of this PR is to prevent the canvas from overlapping with the logarithmic checkbox on narrower viewports. Is that right?

Screenshot from 2025-08-20 11-16-09

This PR fixes that, but I must say I really liked having the external legend / table next to the chart when viewing the sub-tool directly! It is odd I think to have the legend in the form of a table and it be beneath the chart.

Maybe this could all be simplified if we use flexbox? In particular, since the logarithmic checkbox is solely to do with the chart, it should perhaps be in the same container element.

If the above doesn't sound feasible or not worth the effort, feel free to self-merge. This PR is still an improvement :)

agoujot:
There we go; did that. Flexbox indeed made the whole lot simpler (plus, no floating, so no more RTL issues). Merging.

PR #539, opened 2025-06-03
Fix error messages and codes for timeouts

agoujot:
Split messages depending on whether we or the server stopped the request.
Fix a helper function returning 500 in all cases.
Fix the timeout being stored where the status code should go.

Bug: T226307

### DIFF
--- a/src/EventSubscriber/ExceptionListener.php
+++ b/src/EventSubscriber/ExceptionListener.php
@@ -144,7 +144,7 @@ class ExceptionListener
 
         return new Response(
             $this->templateEngine->render('bundles/TwigBundle/Exception/error.html.twig', [
-                'status_code' => Response::HTTP_INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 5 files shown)


src/EventSubscriber/ExceptionListener.php
```diff
@@ -144,7 +144,7 @@ private function getTwigErrorResponse(Throwable $exception): Response
 
         return new Response(
             $this->templateEngine->render('bundles/TwigBundle/Exception/error.html.twig', [
-                'status_code' => Response::HTTP_INTERNAL_SERVER_ERROR,
+                'status_code' => $exception->getCode(),
```
MusikAnimal, line 147: All this time… lol 🤦

MusikAnimal (approved):
Thank you!!! :100:

These errors are hard to test, as you know, but this looks like it will work. Merging!

PR #544, opened 2025-06-06
PageHistory: add subpages count to generalstats

agoujot:
Bug: T395765


templates/pageInfo/result.html.twig
```diff
@@ -158,6 +158,16 @@
...
+                    {% if ai.subpageCount > 0 %}
+                        <tr>
+                            <td>{{ msg('subpages') }}</td>
+                            <td>
+                                {{
+                                    wiki.pageLinkRaw("Special:PrefixIndex/" ~ page.title, project, ai.subpageCount|num_format)
+                                }}
+                            </td>
+                        </tr>
+                    {% endif %}
```
MusikAnimal: I'm thinking this might be better placed under the "Links" section. Even though it's not technically referring to a links per se, this stat like the other "Links" stats has to do with related pages. What do you think?

I'd suggest putting it at the bottom and giving the `<td>`s a `class="stat-list--new-group"` so it's visually separated from Categories, Files and Templates.

agoujot: Fine by me.

PR #545, opened 2025-06-08
PageInfo: fix post-import interwiki links

agoujot:
Given MW auto redirects valid interwiki prefixes, the solution ended up being very easy.

Bug: T330941


MusikAnimal (approved):
Works great for most places! The interwiki prefixes are still there in the charts and legends, and maybe other places too (JSON output should remain with the prefix, I think), but this is certainly a great improvement. Thank you!

PR #547, opened 2025-06-10
TopEdits: limit per ns in php not sql, assigment errorring in prod

agoujot:
Shouldn't be slower, given we already query all the data. Fixes a prod error happening at commons and other projects.

I don't know why the error happens, but it's related to the assignments; at least this gives a fix while we investigate.

Bug: T396423


MusikAnimal:
I have cherry-picked this in prod. Seems to work great. Thank you :D

PR #549, opened 2025-06-21
PageInfo: wait onload for script

agoujot:
Perhaps for some very simple PRs like this I should merge them myself? IDK.

### DIFF
--- a/templates/pageInfo/result.html.twig
+++ b/templates/pageInfo/result.html.twig
@@ -455,7 +455,9 @@
         {# ======================== AUTHORSHIP ======================== #}
         {% if showAuthorship %}
             <script>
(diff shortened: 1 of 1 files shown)


MusikAnimal:
> Perhaps for some very simple PRs like this I should merge them myself? IDK.

Yes! For very simple ones you're certain are good, feel free to self-merge. I self-merge, and you've proven yourself. Also I can't really stop you, lol.

Note that https://xtools-dev.wmcloud.org gets updated every 10 minutes with `main`, so you will be able to test there before we do a prod release. I will however be using the dev server for testing Anubis. Over the next week or so, it may not be a stable environment.

PR #551, opened 2025-06-27
Add ProofreadPage quality support and cover all Model

agoujot:
In common
-----
(Sorry if I use WS jargon, feel free to ask.)
 - Add an API call to get the names of the PRP categories (avoids retranslation, and respects possible onwiki customisation).
 - In getMetadata, also gets the canonical names of namespaces. The best way we have at hand of checking which ID the PRP Page namespace has is that it's the one with the canonical name "Page".
 - Some styling copied from a template and adjusted for sufficient contrast (including dark mode)

EditCounter
-----
 - Get the tags of the last 5k revisions (also done by #532).
 - Add a little pie chart at the end of general_stats. Looks like [this](https://github.com/user-attachments/assets/673514ce-913b-4fdc-88ea-9469fddbc311).

PagesCreated
-----
 - in getPagesCreatedInnerSql, remove the DISTINCT as I'm having trouble seeing why it's there and it makes the query significantly slower (when you add in the PRP stuff)
 - For pages in the right namespace, join on `page_props` with `pp_propname = "proofread_page_quality_level"`
 - Add a column to the table of that namespace. Looks like [this](https://github.com/user-attachments/assets/11ccef55-0726-4ab4-a3e9-358a3d9ace4a).
 - In countPagesCreated count the number of pages of each status.
 - Add a pie chart next to that table. Looks like [this](https://github.com/user-attachments/assets/44722bf2-ee2e-4511-a657-5beb9de6c28f).

TopEdits
-----
 - For convenience, clean up the mess of `getTopEditsAllNamespaces`. Change from a 3-level query with custom row numbering with assignments and a random join at the end for the title (?) to a 2-level query with the built-in `ROW_NUMBER` (the one that caused trouble when it reserved the name recently).
 - Still for convenience, Factorise the PA conditions.
 - For pages in the right namespace, join on `page_props` like in PC
 - Add a column in the same way.

Don't make a chart, though, because it implies being done over all the data and we can't do that for TopEdits (I think?). At any rate, right now all we do is count 'em; whereas for PC we already make some other stats.

Tests
-----
I'm also adding many, many tests. That has led me to refactorise some code, fix a few bugs here and there, make a few private methods public, etc. Feel free to ask for the reason behind specific changes. Goal would be 100% on model and some other stuff.

Bug: T173012


agoujot:
Don't merge: just found out we got a `page_props` for this. Completely changes the field of possibilities.

agoujot:
Oh, and I also should add test whenever I'm finished, shouldn't I? This PR is going to be large enough to reduce coverage by multiple percentage points if I don't xD.

agoujot:
Phew, now mergeable with main. The changes in the last few months mean we're not at 100% model anymore; I'll try and put Model & Helper (& maybe Repository &/| EventSuscriber) to 100%. It's only 46 (not counting repo & events) lines to cover.

PR #558, opened 2025-07-19
PageInfo: add tag support: mobile, visual, automated

agoujot:
- Get tags in getRevisionsStmt in Repo
 - Give Edit support for tags (as list of tag names)
 - Count visual and mobile edits
 - Add two lines on the leftmost table of generalstats that give those two numbers
 - Make AutomatedEditsHelper's getTool take an optional tag list into account.

To test the latter, go with https://localhost:8000/pageinfo/enwikisource/File:A%20Beginners%20First%20Course%20In%20Cutting.djvu . There the only auto edit is a PAWS edit, which doesn't have a regex.
I've lumped here two tasks that else would have had much too much code in common.

Bug: T356707
Bug: T340254


src/Model/Edit.php
```diff
@@ -461,6 +467,16 @@ public function isAnon(Project $project): ?bool
         return $this->getUser() ? $this->getUser()->isAnon($project) : null;
     }
 
+    /**
+     * List of tag names for the edit.
+     * Only filled in in PageInfo.
```
MusikAnimal:
```suggestion
     * Only filled in by PageInfo.
```

src/Model/Edit.php
```diff
@@ -90,6 +93,9 @@ public function __construct(EditRepository $repository, UserRepository $userRepo
         $this->lengthChange = isset($attrs['length_change']) ? (int)$attrs['length_change'] : null;
         $this->comment = $attrs['comment'] ?? '';
 
+        // Had to be JSON to put multiple values in 1 column.
+        $this->tags = JSON_decode($attrs['tags'] ?? '[]');
```
MusikAnimal:
```suggestion
        $this->tags = json_decode($attrs['tags'] ?? '[]');
```

src/Helper/AutomatedEditsHelper.php
```diff
@@ -37,16 +37,17 @@ public function __construct(RequestStack $requestStack, CacheItemPoolInterface $
...
+     * @param string[]|null $tags
      * @return string[]|null Tool entry including key for 'name', or false if nothing was found
      */
-    public function getTool(string $summary, Project $project): ?array
+    public function getTool(string $summary, Project $project, ?array $tags = []): ?array
     {
         foreach ($this->getTools($project) as $tool => $values) {
-            if (isset($values['regex']) && preg_match('/'.$values['regex'].'/', $summary)) {
+            if ((isset($values['regex']) && preg_match('/'.$values['regex'].'/', $summary)) ||
+                (isset($values['tags']) && count(array_intersect($values['tags'], $tags)) > 0)) {
```
MusikAnimal: What happens if `$tags` is `null`? It looks like this would error out. Does it need to be nullable?

Nitpick:
```suggestion
                (isset($values['tags']) && count(array_intersect($values['tags'], $tags)) > 0))
            {
```

templates/pageInfo/result.html.twig
```diff
@@ -310,6 +310,22 @@
...
                         </tr>
+                        <tr>
+                            <td class="stat-list--new-group">{{ msg('mobile-edits') }}</td>
+                            <td class="stat-list--new-group">
+                                {{ ai.mobileCount|num_format }}
+                                &middot;
+                                ({{ ai.mobileCount|percent_format(ai.numRevisionsProcessed) }})
+                            </td>
+                        </tr>
+                        <tr>
```
MusikAnimal: Ideally we'd only show "Visual edits" if VisualEditor is installed.

templates/pageInfo/result.html.twig
```diff
@@ -310,6 +310,22 @@
                                 ({{ ai.topTenCount|percent_format(ai.numRevisionsProcessed) }})
                             </td>
                         </tr>
+                        <tr>
+                            <td class="stat-list--new-group">{{ msg('mobile-edits') }}</td>
+                            <td class="stat-list--new-group">
+                                {{ ai.mobileCount|num_format }}
```
MusikAnimal: We should link to the results, since we can :smiley: `wiki.pageHistLink()` doesn't take a `queryString`. You could fix that or use something like:
```suggestion
                                {{ wiki.pageLinkRaw( page.title, project, ai.mobileCount|num_format, 'action=history&tagfilter=mobile+edit' }}
```

Same for the visual edit count.

templates/pageInfo/result.html.twig
```diff
@@ -310,6 +310,22 @@
                                 ({{ ai.topTenCount|percent_format(ai.numRevisionsProcessed) }})
                             </td>
                         </tr>
+                        <tr>
+                            <td class="stat-list--new-group">{{ msg('mobile-edits') }}</td>
```
MusikAnimal: "Mobile edits" can be ambiguous. Perhaps we could add a `title` attribute with a `mobile-edits-tooltip` message?

MusikAnimal:
Sorry for the nitpicks

src/Helper/AutomatedEditsHelper.php
agoujot: TopEdits also uses this, though isAutomated(); and TopEdits doesn't include the tags. However, it's true that the default value of [] is not null, and so it doesn't need to be nullable.

agoujot:
I think I've addressed everything.

templates/pageInfo/result.html.twig
```diff
@@ -310,6 +310,28 @@
                                 ({{ ai.topTenCount|percent_format(ai.numRevisionsProcessed) }})
                             </td>
                         </tr>
+                        <tr>
+                            <td class="stat-list--new-group">
+                                <span class="mobile-edits-tooltip" title="{{ msg('mobile-edits-tooltip') }}">
+                                    {{ msg('mobile-edits') }}
+                                </span>
+                            </td>
```
MusikAnimal: Twig doesn't automatically remove spacing between tags. In this case the newline and spaces after the `</span>` cause a visible gap in the label. The solution I used for similar labels was:

```suggestion
                            {% apply spaceless %}
                                <td class="stat-list--new-group">
                                    <span class="help-text" title="{{ msg('mobile-edits-tooltip') }}">{{ msg('mobile-edits') }}</span>
                                </td>
                            {% endapply %}
```

Note also we're using the preexisting `help-text` CSS class instead of a new one.

assets/css/pageinfo.scss
```diff
@@ -71,6 +71,11 @@
         margin-top: 5px;
     }
 
+    .mobile-edits-tooltip {
+        cursor: help;
+        border-bottom:thin dotted gray;
+    }
```
MusikAnimal: Seems redundant to the existing `.help-text` class that's in application.scss

MusikAnimal (changes requested):
Just a tiny CSS issue, otherwise looks good to me :+1:

assets/css/pageinfo.scss
agoujot: Indeed! Hadn't noticed.

templates/pageInfo/result.html.twig
agoujot: Actioned.

PR #559, opened 2025-07-21
EditCounter: add an explanation for impossible rights logs

agoujot:
Namely, when the removal of a right is logged without its addition ever having been logged. Use [this](https://xtools.wmcloud.org/ec-rightschanges/en.wikipedia.org/Altenmann) to test: it has a -sysop but not a +sysop.

Bug: T396285


src/Model/UserRights.php
```diff
@@ -305,6 +313,13 @@ private function processRightsChanges(array $logData): array
                 $added = array_values($added);
                 $removed = array_values($removed);
             }
+
+            if (array_diff($removed, $tempRights)) {
+                // Then we're removing something which isn't there.
+                $this->impossibleLogs = true;
+            }
+            $tempRights = array_diff($tempRights, $removed);
+            $tempRights = array_merge($tempRights, $added);
```
MusikAnimal: What do these two lines do?

agoujot: It just updates `tempRights`; we remove the ones that were removed from our temporary list, and add in those that were removed. (I think array_diff and array_merge are the right functions for this. array_diff returns the values of the first argument that aren't the second; i.e. those that were not removed. array_merge just puts the values end to end, but that shouldn't be a problem, given you can't +group if user already has it.)

Still, probably deserved a comment or two. Done.

PR #560, opened 2025-07-23
EditCounter: prevent edit size counts from counting each other.

agoujot:
This would also have been solved by a merge of #532 or #551 ; as these implement a cleaner separation between the raw data and the counts we're making. Until we get around to doing that, this is a simpler and smaller fix.

Bug: T400293


agoujot:
(The root problem is that I had assumed that the PHP array counting/filtering functions only read numeric keys. I think I thought that because it does that in JS.)

PR #561, opened 2025-07-25
Group by rev page

eranroz:
Proposal to group by rev page, to address issues where 2 revisions with parent rev 0 belong to same page.


agoujot:
Thanks for contributing! I tested; it works. Also passes CI. I don't think this will have noticeable impact on performance. Merging.

PR #563, opened 2025-08-03
AutoEdits: don't count tools with contribs=true as automated

agoujot:
Tools with the contribs flag set (eg CXT) are not considered really automated; to be precise, we show them in the non-automated list. This can make the UI confusing, with in extreme cases "0 non-automated edits" but edits in the non-automated list. This PR makes them consistently count as non-automated; thought they still appear in the tools table, renamed from "(Semi-)automated tools" to "Tool counts". (I left the `auto-edits` message because it's also used in other places.)

Bug: T257723


src/Repository/AutoEditsRepository.php
```diff
@@ -585,7 +585,7 @@ private function getToolRegexAndTags(
         }
 
         foreach (array_values($tools) as $values) {
-            if ($nonAutoEdits && isset($values['contribs'])) {
+            if (isset($values['contribs'])) { // shown in the tool list but not counted as automated
```
MusikAnimal, line 586: The doc block says the `$nonAutoEdits` argument can be used to exclude tools with the `contribs` flag, but it doesn't do anything anymore. All the callers appear to pass in `false`, so maybe we should just remove the `$nonAutoEdits` argument and update the callers.

MusikAnimal (changes requested):
Just one tech-debt-related request.

src/Repository/AutoEditsRepository.php
agoujot: Normally done.

PR #567, opened 2025-08-07
Sign Action API requests on behalf of user when logged in

agoujot:
<Among others, this means that users logging in to XTools will have access to restricted data they (but not everyone) have access to onwiki (eg page watcher count < 30).

Changes (all in Repository/Repository.php):
 - inject requestStack
 - in executeApiRequest, if the user is logged in, use the oauth client and access token in session to make a signed request

Note that passing along a requestStack is optional, so we don't sign in *all* repositories; it's opt-in. Currently it's UserRepo, plus all of those that just directly extended without a custom __construct (which includes PageRepo).

Bug: T401360


agoujot:
tested,, works: 
image

src/Repository/Repository.php
```diff
@@ -190,13 +198,36 @@ protected function getDbList(): array
...
+            }
+            if ($session && $session->get('logged_in_user')) {
+                $oauthClient = $session->get('oauth_client');
+                $queryString = http_build_query($fullParams);
+                $requestUrl = $project->getApiUrl() . '?' . $queryString;
+                $body = $oauthClient->makeOAuthCall(
+                    $session->get('oauth_access_token'),
+                    $requestUrl
+                );
+                $this->logger->error($body);
```
MusikAnimal: Either change this to `$this->logger->debug()` or `->info()`, or remove it. Logging as an error (when it's not an error) will result in a lot of log spam.

src/Repository/Repository.php
```diff
@@ -190,13 +198,36 @@ protected function getDbList(): array
...
+                return json_decode($body, true);
+            } else { // Not logged in, default to a not-logged-in query
+                $req = $this->guzzle->request(
+                    'GET',
+                    $project->getApiUrl(),
+                    ['query' => $fullParams]
+                );
+                $body = $req->getBody()->getContents();
+                $this->logger->error("no session");
```
MusikAnimal: Same here. We shouldn't log as an error give it is not treated as an error.

MusikAnimal (changes requested):
Works great! Just one issue as far as I can tell.

agoujot:
Duh! that was debug; not really intended to log that. I always use logger->error when testing, because it's the only one I remember that only needs a string as param.

src/Repository/Repository.php
```diff
@@ -190,13 +198,34 @@ protected function getDbList(): array
...
+            } else { // Not logged in, default to a not-logged-in query
+                $req = $this->guzzle->request(
+                    'GET',
+                    $project->getApiUrl(),
+                    ['query' => $fullParams]
+                );
+                $body = $req->getBody()->getContents();
+                return json_decode($body, true);
+            }
+        } catch (ConnectException|ServerException|OAuthException $e) {
```
MusikAnimal, line 228:
```suggestion
        } catch (ConnectException|ServerException $e) {
```

I'm very excited to get this out, but I'm a tiny bit worried something could awry with OAuth. If we catch the exception here, the BadGatewayException will just blame it on Wikimedia, and we [don't even get an email](https://github.com/x-tools/xtools/blob/e0961fdf10d36ce8cf1252ec27d5d76c7483cb99/config/packages/prod/monolog.yaml#L25). It may very well be a Wikimedia issue, but before we wholesale sign all API requests, I think I'd prefer to be made aware of any OAuthExceptions so we can monitor for any issues. If we see there are in fact continually upstream issues that are beyond our control, we start catching the exception again.

MusikAnimal (changes requested):
Sorry, just one last thing!

src/Repository/Repository.php
agoujot: Per [the doc](https://doc.wikimedia.org/mediawiki-oauthclient-php/master/classMediaWiki_1_1OAuthClient_1_1Client.html#a520126de0aeb2efb17bd072fdd83fcd0), makeOAuthCall throws that exception on curl errors. I don't even know if they throw that when the login fails (why would it fail?).

Do we really want an email for each curl failure?

Oh, and also: again, this doesn't sign *every* api request. It only does so for the repositories which a) don't have a custom __construct, and b) those that explicitly pass the request stack to the parent __construct.

MusikAnimal: Come to find out, `OAuthException` is from the [PHP OAuth extension](https://www.php.net/manual/en/class.oauthexception.php), which `mediawiki/oauthclient` doesn't even use! So we won't ever see that exception. It looks like the lib uses a generic `Exception` (boo!). Catching that is too broad.

I'd suggest removing the `OAuthException` and _if_ we see there are a lot of OAuth-releated `Exception`s that are beyond our control, we'll refactor this code to try/catch `Exception`s only around the necessary bits, so that any other Exception gets thrown normally as it should.

agoujot: I'm not catching `\OAuthException`. `OAuthException` is the alias I gave to `MediaWiki\OAuthClient\Exception` ([doc](https://doc.wikimedia.org/mediawiki-oauthclient-php/master/classMediaWiki_1_1OAuthClient_1_1Exception.html)). If I'm understanding correctly the namespacing here (which I perhaps am not given PHP is not my main), they're throwing their own exception which is called `Exception` (which is a bad idea) but it's namespaced under OAuthClient. So, I think that here the `|OAuthException` only catches their specific `Exception`.

agoujot: Let's take a look at the source, shall we? [Client.php](https://github.com/wikimedia/mediawiki-oauthclient-php/blob/master/src/Client.php) is `namespace MediaWiki\OAuthClient;`'d and `throw new Exception`s; I think that means that it throws `MediaWiki\OAuthClient\Exception` (which is defined in [Exception.php](https://github.com/wikimedia/mediawiki-oauthclient-php/blob/master/src/Exception.php) as just a wrapper for a generic `Exception`). So my `OAuthException` (which maybe should be renamed) defined by `use MediaWiki\OAuthClient\Exception as OAuthException` should be the class they throw, no more and no less. Is that correct?

MusikAnimal: Yes, you are correct! I missed that `Exception` was namespaced, and I see you've got the `use MediaWiki\OAuthClient\Exception as OAuthException` in this file.

Alright, I feel better now :smile: Sorry for making you explain everything. Merging!

MusikAnimal (approved):

PR #568, opened 2025-08-09
Check replicas are there, when meta_p.wiki says so

agoujot:
In a nutshell: look up in information_schema if we actually have any tables for those projects.

Bug: T322466


agoujot:
Just realised we do in fact have access to `information_schema.tables`.

agoujot:
I have a working cleaner solution with `information_schema`. For speed concerns (notably EC calling a lot of ->exists() for different projects), these checks should be run in one batch; which can be easily done knowing the list of projects to check.
For tools that can only use one project (most) it's straightforward.
EC pulls all of its multi-project capacity straight from GC.
GC can normally be accomodated by adding a method to Project.php to load the existence of _all_ projects, similar to how `getAll` works.

PR #571, opened 2025-08-14
I18nHelper: validate language codes, falling back to English

MusikAnimal:
PHP 8.1+ is more strict with language codes. This commit guards against
a "Found unconstructed IntlDateFormatter" by validating the language
code.

Bump krinkle/intuition to the latest

Add test

Bug: T400929
Bug: T393736

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "doctrine/doctrine-migrations-bundle": "~2.0",
         "eightpoints/guzzle-bundle": "^8.0",
         "jms/serializer-bundle": "^3.4",
-        "krinkle/intuition": "^1.0",
+        "krinkle/intuition": "^2.3",
         "mediawiki/oauthclient": "^1.2",
         "nelmio/api-doc-bundle": "~4.11",
         "nelmio/cors-bundle": "^2.3",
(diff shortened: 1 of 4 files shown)


agoujot:
Note: this is effectively a cleaner way to do #527, and therefore should also solve T393736. The one thing that 527 has and this does not have is flashes, to inform the user of what went wrong.

MusikAnimal:
> Note: this is effectively a cleaner way to do #527, and therefore should also solve T393736. The one thing that 527 has and this does not have is flashes, to inform the user of what went wrong.

Ah! Sorry for stepping on toes. Personally I don't think we need a flash message. MediaWiki will silently ignore and fallback to the site language. We use English because there is no site language. So unless there are strong feelings, maybe we can do without the flash message? If we do want it, we should use dependency injection to get a `FlashBagInterface` service as the means to make it available to I18nHelper.

For now I will fix the test failure (PHPCS violation), and if you want we can tackle the flash message in a separate PR? Hope that's OK :)

agoujot:
Fine by me. No problems of toe-stepping, was just noting similarity.

PR #574, opened 2025-08-19
TopEdits: add summary section, include total edits

MusikAnimal:
This commit began as just a fix for the show/hide links next to the
namespace headings when viewing TopEdits from an EditCounter result.
Then I noticed TopEdits::getNumPagesAnyNamespace() was being called 22+
times (for each namespace), which is far too inefficient. I think the
more meaningful statistic for the 'all' namespaces view is unique pages
edited across all the namespaces. Conveniently, this query is already
ran as is necessary for pagination, so there's no cost added.

Rename TopEditsRepository::countEditsNamespace() to what it actually is,
::countPagesNamespace(), and give it proper support for 'all'
namespaces. Thus, we no longer need TopEdits::getNumPagesAnyNamespace().

One thing led to another, and we are now showing the familiar "Summary"
section on TopEdits namespace result pages. This also surfaces the
start/end date, which should be there to begin with for clarity and
consistency with other result pages.

Rework TopEdits::getNumTopEdits() to do more what it sounds like: for
TopEdits namespace, it gives the count of total edits. Before this
method was only used for TopEdits per page. The "Total edits" for the
given conditions (namespace, date range) is now exposed. In my tests it
can add a few seconds for extremely prolific users, but usually it's
quite fast.

Finally, fix pagination links to use the start and end dates when
applicable.

Bug: T218531

### DIFF
--- a/templates/topedits/result_namespace.html.twig
+++ b/templates/topedits/result_namespace.html.twig
@@ -14,6 +14,40 @@
             {% if project.userHasOptedIn(user) and te.topEdits|length > 2 %}
                 {{ layout.nsToc(project, te.topEdits|keys) }}
             {% endif %}
+
+            {% set content %}
+                <div class="col-lg-6 stat-list clearfix">
+                    <table class="table"><tbody>
+                        <tr>
+                            <td>{{ msg('namespace') }}</td>
+                            <td>
+                                {{ nsName(te.namespace, project.namespaces) }}
+                            </td>
+                        </tr>
+                        {% if te.start is not empty %}
+                            <tr>
+                                <td>{{ msg('start') }}</td>
+                                <td>{{ te.startDate }}</td>
+                            </tr>
+                        {% endif %}
+                        {% if te.end is not empty %}
+                            <tr>
+                                <td>{{ msg('end') }}</td>
+                                <td>{{ te.endDate }}</td>
+                            </tr>
+                        {% endif %}
+                        <tr>
+                            <td class="stat-list--new-group">{{ msg('total-edits') }}</td>
+                            <td class="stat-list--new-group">{{ te.numTopEdits|num_format }}</td>
+                        </tr>
+                        <tr>
+                            <td>{{ msg('unique-pages-edited') }}</td>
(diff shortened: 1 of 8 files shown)


templates/topedits/result_namespace.html.twig
```diff
@@ -14,6 +14,40 @@
...
+                        {% if te.start is not empty %}
+                            <tr>
+                                <td>{{ msg('start') }}</td>
+                                <td>{{ te.startDate }}</td>
+                            </tr>
+                        {% endif %}
+                        {% if te.end is not empty %}
+                            <tr>
+                                <td>{{ msg('end') }}</td>
+                                <td class="xt-test--end-date">{{ te.endDate }}</td>
```
agoujot: What's `xt-test--end-date` for? It has no added CSS anywhere. Grepping, it only appears in:
```
templates/categoryEdits/result.html.twig
templates/topedits/result_page.html.twig
templates/topedits/result_namespace.html.twig
templates/pages/result.html.twig
templates/autoEdits/result.html.twig
```
(Each time being just added to a `td`.)

Also, `templates/editCounter/general_stats.html.twig` uses `xt-test--total-edits`, that doesn't have any styling either.

If we really use these classes, they probably should be moved to clearer names that doesn't contain `test`. (Or is that an acronym for something? :thinking: IDK; only common point between the uses is appearing in the summary box.) And also, why would we specifically add styling to the end date but not the start date? If we aren't doing anything with them, might as well remove them, no?

agoujot (changes requested):
Looks good to me, mostly. Tested around, works. There's just one thing.

templates/topedits/result_namespace.html.twig
MusikAnimal: Nice catch! I think the `xt-test--` classes were used in integration tests, say to verify the displayed end date is the expected. I don't remember ever actually using them… so maybe we were planning to and just never wrote those tests, or they were removed and we forgot to remove the test classes. I dunno. I've removed all of them with the latest push.

PR #576, opened 2025-08-19
Upgrade to Symfony 6.4 and employ new PHP 8.2 features

MusikAnimal:
Note: This includes #594. Also make sure you run `composer install` before testing!

In addition to bringing us up-to-date with the latest LTS release,
this fixes the annoying issue with local envs where the cache directory
had to be manually cleared. Symfony 6 brings a refreshed debug toolbar
among other benefits as well.

Employed PHP features now that we can use 8.2:
- We now use native PHP method attributes instead of annotations for
  routing. Other annotations can likely be moved to attributes at some
  point, too.
- Use unioned types in method signatures where possible, and if the
  resulting doc block adds no information, remove it
- Use constructor property promotion
- Use str_starts_with() where possible
- Use nullsafe operations in conjunction with null coalesce
- Remove unused variables from signature of catch statements
    
Bug: T400929


MusikAnimal:
@agoujot I'm not asking for a thorough review here (it's so big of a PR that GitHub won't even let you browse but one file at a time!), but if you could test this locally to make sure things generally look good, etc., that'd be helpful! I thought it would be good to get this out before my sabbatical (i.e. the annoying issue with caching on local envs).

MusikAnimal:
If we get this merged in time, I will also help resolve conflicts in your other PRs, if you'd like. They shouldn't be _too_ terrible I imagine, but there will certainly be some conflicts.

agoujot (approved):
LGTM? Played around some, everything seems to still be working.

PR #580, opened 2025-09-01
DefaultController: catch oauth exceptions during login

agoujot:
Nothing else to say?


agoujot:
D'you mind taking a look? We've been getting quite many failmails about this.

PR #593, opened 2025-12-21
GC, PC: use - in pagination to prevent slash collapsing

agoujot:
Hopefully we should get a better way out, but meanwhile.... here we are.

(note: CI appears to be failing again: tests pass locally)

Bug: T403566


MusikAnimal:
Re CI failure: It looks like assets were generated with the wrong version of NPM/Node possibly, or you forgot to commit them?

Anyway, once I finally found the time, I was able to fix the core issue in prod: https://phabricator.wikimedia.org/T403566#11485863 Sorry it took me so long to get back to giving it another try.

agoujot:
This was more of an emergency fix for the two tools concerned, and it's not a proper switch to always using -s. Now that the urgent issue is gone, we've got some more time to find something better.

PR #594, opened 2025-12-28
TopEdits: fix wikitext output for top projects and simplify queries

MusikAnimal:
I believe I even requested to combine into one query, but inadvertently this put it in the API output as a serialized JSON string. That could be fixed, but this wasn't intended to be there to begin with. Also, combining the queries is only applicable when there is one page of results, which is usually only the case for less prolific users - in which case doing a separate query is negligible. So this just simplifies things.


MusikAnimal:
Included as part of #576

PR #602, opened 2026-01-03
Fix a typo

NguoiDungKhongDinhDanh:
[The last commit](https://github.com/x-tools/xtools/commit/08c58c4036022c78542c51116395c9dfe62eecb6) renamed `loadinganimation` to `loadingAnimation` without also changing the reference to that variable.

PR #603, opened 2026-01-03
Fix variable capitalization

pppery:
Bug: T413692


agoujot:
Done already with #602 .

PR #615, opened 2026-01-21
Escape uselang param when checking against project name

samwilson:
To avoid syntax errors etc.

Bug: T414901


samwilson:
Merging as this was reviewed on the Phab task.

